PR #2, opened 2018-11-30
IndyCrypto history for URSA

jovfer:
This PR is append for @mikelodder7 effort about migration https://github.com/hyperledger/indy-crypto to URSA.


mikelodder7:
I actually think let’s just use this to start. That’s @jovfer

dcmiddle:
Is the intent of this PR to bring over the history correlated with crypto-lib/libhl-crypto prior to the point that we worked on the joint signing API?
I see a wrappers directory in this PR that is not part of crypto-lib. Is that intentional?

hartm:
@dcmiddle Yes, the point is to bring over the history.

mikelodder7:
Yes I couldn’t get the history the first time but have worked it out now

dcmiddle:
Let me rephrase... where do the two histories correlate and can you clarify why there's an additional directory structure that doesn't exist in crypto-lib?
i.e. are you trying to replace this commit https://github.com/hyperledger-labs/crypto-lib/commit/7d271e9bf62859720054c32d74119090a77a7c72 with the original indy history and then lay the rest of the crypto-lib history over it? or is this at a later point right before we added the common signing API... ?

mikelodder7:
> Let me rephrase... where do the two histories correlate and can you clarify why there's an additional directory structure that doesn't exist in crypto-lib?
> i.e. are you trying to replace this commit [hyperledger-labs/crypto-lib@7d271e9](https://github.com/hyperledger-labs/crypto-lib/commit/7d271e9bf62859720054c32d74119090a77a7c72) with the original indy history and then lay the rest of the crypto-lib history over it? or is this at a later point right before we added the common signing API... ?
Yes so this would be the starting point for ursa then take existing crypt-lib and lay it on top of it. 
crypto-lib started just like this then I began to modify it to have crypto-lib looks today. So this is the first PR. The next one would take crypto-lib as is and merge it with this. The third one would take z-mix and merge it.

dcmiddle:
Ok that works for me. I added Manu as a reviewer so we have a second diverse reviewer.

PR #5, opened 2018-12-12
CWE-325: Missing Required Cryptographic Step.

EvertonMelo:
indy-crypto-deps.zip 
hash: f2bcfe3523201d836e6c3cdaf8189aebf1a68a19

Description

The software does not implement a required step in a cryptographic algorithm, resulting in weaker encryption than advertised by that algorithm.
Context

The Phase identifies a point in the software life cycle at which introduction may occur, while the Note provides a typical scenario related to introduction during the given phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ us on [Jira's Rocket.Chat](chat.hyperledger.org) at #indy-sdk to discuss.
 
     * Download the prebuilt dependencies [here](https://repo.sovrin.org/windows/libindy_crypto/deps/)
     * file: indy-crypto-deps.zip 
-    * hash: f2bcfe3523201d836e6c3cdaf8189aebf1a68a19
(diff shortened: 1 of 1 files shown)


mikelodder7:
This looks fine. Once you fix the DCO I can merge it. See the steps for fixing it [here](https://github.com/hyperledger/ursa/pull/5/checks)

dhuseby:
+1

Thanks for updating the docs.  In the future I'd like to see us use a more robust file integrity mechanism.  Something like [minisign](https://jedisct1.github.io/minisign/) or the OpenBSD signify tool.

hartm:
This is, in general, a good change.  But why is the checksum using a SHA-1 hash instead of a more modern (and less broken) hash function?

And +1 to Dave.  We will need a more robust tool in the long run.

mikelodder7:
No longer applicable.

PR #8, opened 2019-01-14
Added rust implementation for BigNumber

mikelodder7:
This PR adds a pure rust big number implementation.
This is the last step to being able to build WASM for the CL module.


libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+
+    pub fn generate_prime(size: usize) -> Result<BigNumber, UrsaCryptoError> {
+        match prime::new(size)?.to_bigint() {
+            Some(bn) => Ok(BigNumber { bn }),
+            None => Err(UrsaCryptoError::InvalidStructure("Unable to generate prime".to_string()))
+        }
+    }
+
+    pub fn generate_safe_prime(size: usize) -> Result<BigNumber, UrsaCryptoError> {
+        match safe_prime::new(size)?.to_bigint() {
```
lovesh:
```
match prime::new(size)?.to_bigint() {
            Some(bn) => Ok(BigNumber { bn }),
            None => Err(UrsaCryptoError::InvalidStructure("Unable to generate prime".to_string()))
        }
```
This code can be encapsulated in a macro that takes a prime generator, i.e `prime` or `safe_prime` thus avoiding code duplication

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            Some(bn) => bn,
+            None => return Err(UrsaCryptoError::InvalidStructure(format!("Invalid number for 'end': {:?}", end)))
+        };
+
+        let bits = (&end -  &start).bits();
+        let mask = (BigUint::from(3u8) << (bits - 2)) | BigUint::one();
+
+        loop {
+            res = rng.gen_biguint_range(&start, &end);
+            res |= &mask;
```
lovesh: A comment explaning the need of applying the mask would be very helpful

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        } else {
+            match self.bn.to_biguint() {
+                Some(bn) => Ok(prime::check(&bn)),
+                None => Err(UrsaCryptoError::InvalidStructure("An error in is_prime".to_string()))
+            }
+        }
+    }
+
+    pub fn is_safe_prime(&self, _ctx: Option<&mut BigNumberContext>) -> Result<bool, UrsaCryptoError> {
+        if self.bn < BigInt::zero() {
```
lovesh: Why not use `is_negative()`?

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            Ok(false)
+        } else {
+            match self.bn.to_biguint() {
```
lovesh: Can't this logic be abstracted in a macro taking `prime` or `safe_prime` and an error message as an argument?

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+
+    pub fn from_dec(dec: &str) -> Result<BigNumber, UrsaCryptoError> {
+        Ok(BigNumber{ bn: BigInt::from_str_radix(dec, 10)? })
+    }
+
+    pub fn from_hex(hex: &str) -> Result<BigNumber, UrsaCryptoError> {
+        Ok(BigNumber{ bn: BigInt::from_str_radix(hex, 16)? })
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> Result<BigNumber, UrsaCryptoError> {
```
lovesh: I assume this is for deserializing a stored `BigInt`. Does this assume that the serialized `BigInt` was positive?

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn to_hex(&self) -> Result<String, UrsaCryptoError> {
+        Ok(self.bn.to_str_radix(16).to_uppercase())
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, UrsaCryptoError> {
+        let (_, res) = self.bn.to_bytes_be();
+        Ok(res)
+    }
+
+    pub fn hash(data: &[u8]) -> Result<Vec<u8>, UrsaCryptoError> {
```
lovesh: Shouldn't we have several methods like `sha2_256_hash`, `sha2_512_hash`, `sha3_256_hash`

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        Ok(self)
+    }
+
+    pub fn div_word(&mut self, w: u32) -> Result<&mut BigNumber, UrsaCryptoError> {
+        self.bn /= w;
+        Ok(self)
+    }
+
+    pub fn mod_exp(&self, a: &BigNumber, b: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        if a.is_negative() {
```
lovesh: `modpow` [panics]https://docs.rs/num-bigint/0.2.2/num_bigint/struct.BigInt.html#method.modpow() if modulus is less than 1 so a check for `b` is needed.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            let a = a.set_negative(false)?;
+            Ok(BigNumber{ bn: res.bn.modpow(&a.bn, &b.bn) })
+        } else {
+            let res = self.bn.modpow(&a.bn, &b.bn);
+            Ok(BigNumber { bn: res })
+        }
+    }
+
+    pub fn modulus(&self, a: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        let res = &self.bn % &a.bn;
```
lovesh: Check needed to ensure `a` > 0

lovesh: I think it makes to add the above check and use this method instead of `%`

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn lshift1(&self) -> Result<BigNumber, UrsaCryptoError> {
+        Ok(BigNumber { bn: &self.bn << 1 })
+    }
+
+    pub fn rshift1(&self) -> Result<BigNumber, UrsaCryptoError> {
+        Ok(BigNumber { bn: &self.bn >> 1 })
+    }
+
+    pub fn rshift(&self, n: i32) -> Result<BigNumber, UrsaCryptoError> {
+        let n = n as usize;
```
lovesh: a negative `n` will result in a panic

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        Ok(BigNumber { bn: &self.bn >> n })
+    }
+
+    pub fn mod_div(&self, b: &BigNumber, p: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        //(a * (1/b mod p) mod p)
+        let res = (&self.bn * b.inverse(p, _ctx)?.bn) % &p.bn;
```
lovesh: `p` can be negative, better to add the negative check in `modulus` and use the method.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn exp(&self, a: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        if self.bn.bits() == 0 {
+            return Ok(BigNumber::default())
+        } else if a.bn.is_one() {
+            return Ok(self.clone()?)
+        }
+
+        match a.bn.to_u64() {
+            Some(num) => Ok(BigNumber { bn: self.bn.pow(num) }),
+            None => Err(UrsaCryptoError::InvalidStructure("'a' cannot be help in u64".to_string()))
```
lovesh: Error message seems odd

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn mod_sub(&self, a: &BigNumber, n: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        let res = (&self.bn - &a.bn) % &n.bn;
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn div(&self, a: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        let res = &self.bn / &a.bn;
```
lovesh: Check for `a` != 0

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        }
+    }
+
+    pub fn inverse(&self, n: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        if n.bn.is_one() ||
+           n.bn.is_zero() {
+            return Err(UrsaCryptoError::InvalidStructure("Invalid modulus".to_string()))
+        }
+
+        // Euclid's extended algorithm, Bèzout coefficient of `n` is not needed
```
lovesh: Why the comment about "Bèzout coefficient"? Do you intend to have separate method when `n` is coprime to `self`

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        let (mut t, mut new_t) = (BigInt::zero(), BigInt::one());
+        let (mut r, mut new_r) = (n.bn.clone(), self.bn.clone());
+
+        while !new_r.is_zero() {
+            let quotient = &r / &new_r;
+            let temp_t = t.clone();
+            let temp_new_t = new_t.clone();
+
+            t = temp_new_t.clone();
+            new_t = temp_t - &quotient * temp_new_t;
```
lovesh: Moving `new_t = temp_t - &quotient * temp_new_t;` above `t = temp_new_t.clone();` should save the clone of `t`, i.e `temp_t`, right?

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        //    if t < 0 then t := t + n
+        //    return t
+        //
+        let (mut t, mut new_t) = (BigInt::zero(), BigInt::one());
+        let (mut r, mut new_r) = (n.bn.clone(), self.bn.clone());
+
+        while !new_r.is_zero() {
+            let quotient = &r / &new_r;
+            let temp_t = t.clone();
+            let temp_new_t = new_t.clone();
```
lovesh: Do you need this clone?

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            return Ok(self.clone()?)
+        }
+
+        match a.bn.to_u64() {
+            Some(num) => Ok(BigNumber { bn: self.bn.pow(num) }),
+            None => Err(UrsaCryptoError::InvalidStructure("'a' cannot be help in u64".to_string()))
+        }
+    }
+
+    pub fn inverse(&self, n: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
```
lovesh: `_ctx` is unused

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            iteration += 1;
+        }
+
+        match res.to_bigint() {
+            Some(bn) => Ok(BigNumber{bn}),
+            None => Err(UrsaCryptoError::InvalidStructure("Unable to generate prime in range".to_string()))
+        }
+    }
+
+    pub fn is_prime(&self, _ctx: Option<&mut BigNumberContext>) -> Result<bool, UrsaCryptoError> {
```
lovesh: `_ctx` is unused.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            Ok(false)
+        } else {
+            match self.bn.to_biguint() {
+                Some(bn) => Ok(prime::check(&bn)),
+                None => Err(UrsaCryptoError::InvalidStructure("An error in is_prime".to_string()))
+            }
+        }
+    }
+
+    pub fn is_safe_prime(&self, _ctx: Option<&mut BigNumberContext>) -> Result<bool, UrsaCryptoError> {
```
lovesh: `_ctx` is unused.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        let res = &self.bn + &a.bn;
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn sub(&self, a: &BigNumber) -> Result<BigNumber, UrsaCryptoError> {
+        let res = &self.bn - &a.bn;
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn sqr(&self, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
```
lovesh: `_ctx` is unused.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        let res = &self.bn * &self.bn;
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn mul(&self, a: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
```
lovesh: `_ctx` is unused.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+        let res = &self.bn * &a.bn;
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn mod_mul(&self, a: &BigNumber, n: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
```
lovesh: `_ctx` is unused.

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn mod_sub(&self, a: &BigNumber, n: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        let res = (&self.bn - &a.bn) % &n.bn;
+        Ok(BigNumber { bn: res })
+    }
+
+    pub fn div(&self, a: &BigNumber, _ctx: Option<&mut BigNumberContext>) -> Result<BigNumber, UrsaCryptoError> {
+        let res = &self.bn / &a.bn;
+        Ok(BigNumber { bn: res })
+    }
+
```
lovesh: `_ctx` is unused in above methods

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,632 @@
...
+            Some(bn) => bn,
+            None => return Err(UrsaCryptoError::InvalidStructure(format!("Invalid number for 'start': {:?}", start)))
+        };
+        let end = match end.bn.to_biguint() {
+            Some(bn) => bn,
+            None => return Err(UrsaCryptoError::InvalidStructure(format!("Invalid number for 'end': {:?}", end)))
+        };
+
+        let bits = (&end -  &start).bits();
+        let mask = (BigUint::from(3u8) << (bits - 2)) | BigUint::one();
```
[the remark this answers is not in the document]
mikelodder7: Big Primes have the highest two bits set and the lowest bits set. The mask sets those bits.

libursa/src/bn/rust.rs
mikelodder7: Yes but like openssl it will handle the negative case too if the number starts with "-"

libursa/src/bn/rust.rs
mikelodder7: Later but not now.

mikelodder7: Ideally we should remove this and call the hash module.

libursa/src/bn/rust.rs
mikelodder7: To be clear about the method. If we didn't have to handle coprime modulus then we could use other tricks modexp(base, modulus - 2, modulus)

libursa/src/bn/rust.rs
```diff
@@ -0,0 +1,696 @@
...
+
+        if start > end {
+            let temp = start;
+            start = end.clone();
+            end = temp;
+        }
+
+        loop {
+            res = rng.gen_biguint_range(&start, &end);
+            res |= BigUint::one();
```
lovesh: `BigUint::one()` will create a new `BigUint` each time, why not call it once outside the loop, assign the result to a variable and use the variable inside the loop?

mikelodder7: When I view the `asm` code using cargo-asm it appears to be optimized and only create it once anyway.

PR #11, opened 2019-01-25
Fix compiling warnings

cam-parra:
Signed-off-by: Cam Parra <caeparra@gmail.com>

### DIFF
--- a/libzmix/src/zkl/mod.rs
+++ b/libzmix/src/zkl/mod.rs
@@ -4,7 +4,7 @@ pub struct Parser {}
 
 #[allow(unused_variables)]
 impl Parser {
-    pub fn parse(proof_spec: &spec::ProofSpec, _witness: &Witness) -> Result<Proof, ZKLError> {
+    pub fn parse(proof_spec: &spec::ProofSpec, witness: &Witness) -> Result<Proof, ZKLError> {
         unimplemented!();
     }
 }
--- a/libzmix/src/zkl/mod.rs
+++ b/libzmix/src/zkl/mod.rs
@@ -3,7 +3,7 @@ pub mod spec;
 pub struct Parser {}
 
 impl Parser {
-    pub fn parse(proof_spec: &spec::ProofSpec, witness: &Witness) -> Result<Proof, ZKLError> {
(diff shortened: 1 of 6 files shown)


libzmix/src/zkl/mod.rs
```diff
@@ -3,7 +3,7 @@ pub mod spec;
 pub struct Parser {}
 
 impl Parser {
-    pub fn parse(proof_spec: &spec::ProofSpec, witness: &Witness) -> Result<Proof, ZKLError> {
+    pub fn parse(_proof_spec: &spec::ProofSpec, _witness: &Witness) -> Result<Proof, ZKLError> {
```
lovesh: Why not use `#[allow(unused_variables)]`. The reason compiler is complaining about the name is that the function is not implemented. What if we had a Parser trait and this was just one of the implementation. We would not want to change arg names.

libzmix/src/zkl/mod.rs
```diff
@@ -21,7 +21,7 @@ impl WitnessBuilder {
 pub struct Proof {}
 
 impl Proof {
-    pub fn verify(&self, proof_spec: &spec::ProofSpec) -> Result<bool, ZKLError> {
+    pub fn verify(&self, _proof_spec: &spec::ProofSpec) -> Result<bool, ZKLError> {
```
lovesh: Same comment as above

libzmix/src/zkl/mod.rs
cam-parra: I agree with this. Making the changes.

libzmix/src/zkl/mod.rs
```diff
@@ -2,8 +2,9 @@ pub mod spec;
 
 pub struct Parser {}
 
+#[allow(unused_variables)]
 impl Parser {
-    pub fn parse(_proof_spec: &spec::ProofSpec, _witness: &Witness) -> Result<Proof, ZKLError> {
+    pub fn parse(proof_spec: &spec::ProofSpec, _witness: &Witness) -> Result<Proof, ZKLError> {
```
lovesh: What about `_witness`?

PR #23, opened 2019-04-16
All updates for indy to integrate ursa

mikelodder7:


libursa/src/errors/mod.rs
```diff
@@ -2,197 +2,217 @@
...
+    InvalidStructure,
+    #[fail(display = "Invalid parameter {}", 0)]
+    InvalidParam(u32),
+    #[fail(display = "IO error")]
+    IOError,
+    // CL errors
+    #[fail(display = "Proof rejected")]
+    ProofRejected,
+    #[fail(display = "Revocation accumulator is full")]
+    RevocationAccumulatorIsFull,
```
brentzundel: `RevocationAccumulatorIsFull` feels out of place with the rest of the new error types, perhaps this could be made more general? (e.g. in the same way `AnoncredsProofRejected` became `ProofRejected`)
`AccumulatorIsFull` would be better, but perhaps even something more generic could work.

libursa/src/errors/mod.rs
```diff
@@ -2,197 +2,217 @@
...
+    #[fail(display = "Invalid revocation id")]
+    InvalidRevocationAccumulatorIndex,
```
brentzundel: Roughly the same comment as above, this feels too specific. Not every accumulator will be used for revocation.

libursa/src/errors/mod.rs
```diff
@@ -2,197 +2,217 @@
...
+    #[fail(display = "Credential revoked")]
+    CredentialRevoked,
```
brentzundel: I wonder if `CredentialRevoked` would be a more useful error if it were `CredentialNotValid` along with a message about why the credential is invalid.

libursa/src/errors/mod.rs
```diff
@@ -2,197 +2,217 @@
...
-            UrsaCryptoError::InvalidParam9(ref description) => description,
-            UrsaCryptoError::InvalidState(ref description) => description,
-            UrsaCryptoError::InvalidStructure(ref description) => description,
-            UrsaCryptoError::IOError(ref err) => err.description(),
-            UrsaCryptoError::AnoncredsRevocationAccumulatorIsFull(ref description) => description,
-            UrsaCryptoError::AnoncredsInvalidRevocationAccumulatorIndex(ref description) => description,
-            UrsaCryptoError::AnoncredsCredentialRevoked(ref description) => description,
-            UrsaCryptoError::AnoncredsProofRejected(ref description) => description,
+#[cfg(feature = "ffi")]
+impl From<UrsaCryptoErrorKind> for ErrorCode {
```
brentzundel: Is this function intended to allow Indy to use Ursa without needing to change Indy? 
If so, is there a plan in the future to modify Indy to respond to Ursa's errors instead of requiring Ursa to do things like this?

libursa/src/errors/mod.rs
mikelodder7: I didn't pick these the indy team did. This was just a matter of syncing recent changes in indy-crypto that weren't merged over.

libursa/src/errors/mod.rs
mikelodder7: Same response from me as well.

libursa/src/errors/mod.rs
mikelodder7: Not at the moment. But the plan is that ursa will follow its own development plan now and Indy will have to adapt to using ursa.

PR #16, opened 2019-02-18
benchmarking

lovesh:
Signed-off-by: lovesh harchandani <lovesh.bond@gmail.com>

PR #24, opened 2019-05-01
Benchmarking (signatures, bls)

rekmarks:
This PR contains performance benchmarks using Criterion.rs for the following `libursa` modules:
- `signatures`
- `bls`

[The readme](https://github.com/rekmarks/ursa/tree/benchmarking/libursa/benches) reports comparisons between Ed25519, Secp256k1, and BLS for key creation, signing, and single-key verification.

My work is trivially compatible with #16. If #16 is merged first, I can add a commit to incorporate it.

This work was completed for a graduate course project at the University of Pennsylvania and following discussions with @mikelodder7.


libursa/benches/bench_wrappers/signatures.rs
```diff
@@ -0,0 +1,204 @@
...
+        let result = scheme.sign(MESSAGE_1, &s);
+        assert!(result.is_ok());
+    }
+
+    pub fn sign_benchmark(c: &mut Criterion) {
+        c.bench_function("secp256k1 signing", |b| b.iter(|| sign()));
+    }
+
+    fn sign_libsecp256k1() {
+        let context = libsecp256k1::Secp256k1::new();
```
lovesh: Can you please provide the benchmark without creating `context` each time?

libursa/benches/bench_wrappers/signatures.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    pub fn sign_libsecp256k1_benchmark(c: &mut Criterion) {
+        c.bench_function("(external) libsecp256k1 signing", |b| b.iter(|| sign_libsecp256k1()));
+    }
+
+    fn sign_openssl() {
+        let pk = hex::hex2bin(PUBLIC_KEY).unwrap();
+        let h = sha2::Sha256::digest(&MESSAGE_1);
+
+        let openssl_group = EcGroup::from_curve_name(Nid::SECP256K1).unwrap();
```
lovesh: Can you please provide the benchmark without creating `ctx` and `openssl_group` each time?

libursa/benches/bench_wrappers/signatures.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert!(result.is_ok());
+    }
+
+    pub fn verify_benchmark(c: &mut Criterion) {
+        c.bench_function("secp256k1 verification", |b| b.iter(|| verify()));
+    }
+
+    fn verify_libsecp256k1() {
+
+        let context = libsecp256k1::Secp256k1::new();
```
lovesh: Same comment about `context`

libursa/benches/bench_wrappers/signatures.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub fn verify_libsecp256k1_benchmark(c: &mut Criterion) {
+        c.bench_function("(external) libsecp256k1 verification", |b| b.iter(|| verify_libsecp256k1()));
+    }
+
+    fn verify_openssl() {
+
+        let h = sha2::Sha256::digest(&MESSAGE_1);
+        let pk = hex::hex2bin(PUBLIC_KEY).unwrap();
+
+        let openssl_group = EcGroup::from_curve_name(Nid::SECP256K1).unwrap();
```
lovesh: Same comment as above about `ctx` and `openssl_group`

libursa/benches/bench_wrappers/signatures.rs
danintel: Memory allocation is usually a big performance problem in crypto implementations, especially on small payloads and in benchmarks. Moving these outside the implementation, or in this case, outside of the loop should help remove that extra noise/overhead.

dcmiddle:
@rekmarks this looks like a great addition to the library!
Under the hood we are using the bitcoin library for secp256k1. It has an expensive initialization. That is hidden in our facade with the intent that the consumer can create a SignatureScheme object and then reuse it as needed without incurring that cost repeatedly.
Unfortunately that API is not well documented at the moment. It looks like you might have adopted the patterns inside the test modules which call directly into the underlying libs sometimes to check compatibility. If you look at the top of the verify test, for example, you can see the Ursa API.
https://github.com/hyperledger/ursa/blob/master/libursa/src/signatures/secp256k1.rs#L356-L359

I would prefer that we setup the benchmarking to test through those interfaces. That way if we are inadvertently adding overhead to the underlying libraries we will be more likely to see it in the bench results.

In terms of results, I expect secp256k1 to be only a little slower than ed25519 (say within 30%). If you see that the performance is factors different then there's probably something wrong with the code or configuration.

PR #28, opened 2019-05-17
Resolve TODO items left as comments in the bls module.

dcmiddle:
Added doc comments and unit tests for bls methods.


libursa/src/bls/mod.rs
```diff
@@ -583,13 +630,55 @@ mod tests {
     }
 
     #[test]
-    fn pop_new_works() {
+    fn ver_key_as_bytes_works() {
```
lovesh: Testing `as_bytes` without `from_bytes` does not seem very useful unless we had some concrete values with which we could do the equality check for byte value and not just byte length. I find `as_bytes` and `from_bytes` combined into one test better.

mikelodder7: Imagine testing serialization and deserialization separately or together. To me, it's a matter of preference, but I do think it's helpful to test them individually. Serialization is always easier and deserialization is usually more complicated because of the additional checks that must be done.

PR #31, opened 2019-06-03
ProofVerifier supports common attributes. Also ensure same blindings are used for common attributes by checking corresponding `m_hat`s

lovesh:


lovesh:
@mikelodder7 Test in both jobs, i.e. Linux and macOS build passed but still the CI shows failure

mikelodder7:
To pass lovesh you need to run cargo fmt then push again.

lovesh:
Thanks @mikelodder7

libursa/src/cl/verifier.rs
```diff
@@ -204,7 +215,7 @@ impl ProofVerifier {
     ///                                      None).unwrap();
     /// assert!(proof_verifier.verify(&proof, &proof_request_nonce).unwrap());
     /// ```
-    pub fn verify(&self, proof: &Proof, nonce: &Nonce) -> UrsaCryptoResult<bool> {
+    pub fn verify(&mut self, proof: &Proof, nonce: &Nonce) -> UrsaCryptoResult<bool> {
```
mikelodder7: Why is self mutable during verify?

libursa/src/cl/verifier.rs
```diff
@@ -45,16 +46,26 @@ impl Verifier {
...
             credentials: Vec::new(),
+            common_attributes: HashMap::new(),
         })
     }
 }
 
 #[derive(Debug)]
 pub struct ProofVerifier {
     credentials: Vec<VerifiableCredential>,
+    common_attributes: HashMap<String, Option<BigNumber>>,
```
mikelodder7: Why is this anything but a Vec<string>?

libursa/src/ffi/cl/verifier.rs
```diff
@@ -5,6 +5,8 @@ use ffi::ErrorCode;
 
 use std::os::raw::c_void;
 
+// TODO: Expose Verifier's add_common_attribute method
+
```
mikelodder7: Why add the TODO? Just implement it.

libursa/src/cl/verifier.rs
lovesh: Because during verify, `common_attributes` are update from `m_hat`s from each subproof.

libursa/src/cl/verifier.rs
lovesh: The verifier's objective is to ensure that common attributes have same `m_hat`s. So when the verifier processes the first sub-proof, it stores the `m_hat` for corresponding common attributes in the hashmap. For the subsequent attributes, it checks that common attributes have same value.

libursa/src/ffi/cl/verifier.rs
lovesh: Ok

libursa/src/cl/verifier.rs
mikelodder7: Okay so why is it Option<BigNumber> and not BigNumber?

lovesh: Because the verifier specifies the attribute names before it starts verifying the proof. So their value is `None` at that time. A simpler but less efficient implementation would be to use a `HashSet<String>` of attribute names which is filled during calls to `add_common_attribute` and a `HashMap<String, BigNumber>` which is filled during call to `verify`

libursa/src/ffi/cl/verifier.rs
lovesh: I have exposed for both Prover and Verifier

libursa/src/ffi/cl/prover.rs
```diff
@@ -945,6 +945,40 @@ pub extern "C" fn ursa_cl_prover_new_proof_builder(
...
+/// Add a common attribute to the proof builder
+///
+/// # Arguments
+/// * `proof_builder` - Reference that contain proof builder instance pointer.
+/// * `attribute_name` - Common attribute's name
+
+#[no_mangle]
+pub extern "C" fn ursa_cl_proof_builder_add_common_attribute(
+    proof_builder: *const c_void,
+    attribute_name: *const c_char,
```
dhuseby: We really need to get away from pointers. We can't preserve a lot of Rust guarantees if the API allows arbitrary data injection.

libursa/src/ffi/cl/prover.rs
```diff
@@ -1391,6 +1425,75 @@ mod tests {
...
+            credential_values,
+            credential_pub_key,
+            credential_issuance_nonce,
+            ptr::null(),
+            ptr::null(),
+            ptr::null(),
+        );
+        let proof_builder = _proof_builder();
+
+        let common_attr_name = string_to_cstring(String::from("master_secret"));
```
dhuseby: This should be converted to a named static constant outside of the function.

```rust
static MASTER_SECRET: &'static str = "master_secret";
```

libursa/src/ffi/cl/verifier.rs
```diff
@@ -45,6 +46,44 @@ pub extern "C" fn ursa_cl_verifier_new_proof_verifier(
...
+/// Add a common attribute to the proof verifier
+///
+/// # Arguments
+/// * `proof_builder` - Reference that contain proof verifier instance pointer.
+/// * `attribute_name` - Common attribute's name
+
+#[no_mangle]
+pub extern "C" fn ursa_cl_proof_verifier_add_common_attribute(
+    proof_verifier: *const c_void,
+    attribute_name: *const c_char,
```
dhuseby: Again, we need to move away from this ASAP and replace it with some kind of serde marshalling so that the types flowing in/out of the Rust side are checked at compile and runtime.

libursa/src/ffi/cl/verifier.rs
```diff
@@ -144,7 +183,7 @@ pub extern "C" fn ursa_cl_proof_verifier_verify(
     check_useful_c_reference!(nonce, Nonce, ErrorCode::CommonInvalidParam3);
     check_useful_c_ptr!(valid_p, ErrorCode::CommonInvalidParam4);
 
-    let proof_verifier = unsafe { Box::from_raw(proof_verifier as *mut ProofVerifier) };
+    let mut proof_verifier = unsafe { Box::from_raw(proof_verifier as *mut ProofVerifier) };
```
dhuseby: This should be:

```
let mut proof_verifier = serde_json::deserialize::<ProofVerifier>(buf);
```
using unsafe is...uh...*unsafe*.

libursa/src/ffi/cl/verifier.rs
```diff
@@ -258,6 +297,94 @@ mod tests {
...
+            credential_signature,
+            proof_building_nonce,
+            credential_values,
+            ptr::null(),
+            ptr::null(),
+        );
+
+        let proof_verifier = _proof_verifier();
+
+        let common_attr_name = string_to_cstring(String::from("master_secret"));
```
dhuseby: convert to static string constant.

libursa/src/ffi/cl/prover.rs
lovesh: Point noted. I was just following the pattern already being followed.

libursa/src/ffi/cl/verifier.rs
lovesh: I am following the pattern being followed in the FFI code in this project.

libursa/src/ffi/cl/verifier.rs
lovesh: Done

libursa/src/ffi/cl/prover.rs
lovesh: Done

libursa/tests/cl.rs
```diff
@@ -1174,7 +1174,7 @@ mod cl_tests {
                 )
                 .unwrap();
 
-            assert!(proof_verifier.verify(&proof, &nonce).unwrap());
+            assert!(!proof_verifier.verify(&proof, &nonce).unwrap());
```
dcmiddle: Why is this test condition flipped?

dcmiddle:
Which paper is this implemented against (specifically regarding the m_hat references).

libursa/tests/cl.rs
lovesh: The proof was created with 2 credentials with different link secrets and hence was expected to fail.

lovesh:
@dcmiddle The one in the repo [here](https://github.com/hyperledger/ursa/blob/master/libursa/docs/AnonCred.pdf)

libursa/tests/cl.rs
dcmiddle: Thanks for finding that defect.
It looks like it's corrected over three unrelated commits in this PR. Could you please organize the fix into one commit?
Also I'm not clear why this test wasn't preventing CI from passing earlier? I don't see an ignore or anything like that.

lovesh: My first 3 commits were for producing a broken test that illustrated the problem (and some comments). My next commit fixed the issue. Then there are some FFI additions and other changes to address the code review. I would suggest to squash all commits into 2, the first illustrates the issue with a failing test and the second with the fix, FFI included.
As to why the CI did not fail earlier, i had not raised the PR until i did the fix and i don't think CI runs on each commit.

lovesh: @dcmiddle Are you fine with the way commits are laid out now?

dcmiddle: Can you please update the commit messages? I think what this PR is about is there was a defect in the anon credentials code, the unit test which should have exposed the problem had the wrong assertion set so that the test would pass, and you fixed the core issue, corrected the test and added a new test. Whether I got that right or not can you please update the commit messages so the actual story is clearer? All this is obvious when you are the person who spent hours/days/weeks on something and it's entirely not obvious to other people, especially contributors a year from now that will try to figure out the code. :)
This post is helpful:
https://chris.beams.io/posts/git-commit/#seven-rules

lovesh: Done

PR #36, opened 2019-06-12
Adding bulletproofs and gadgets

lovesh:


Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = [
   "libursa",
-  "libzmix"
+  "libzmix",
+  "libzmix/bulletproofs_amcl"
```
mikelodder7: Why do we need another workspace?

libzmix/bulletproofs_amcl/src/r1cs/gadgets/set_membership_alt.rs
```diff
@@ -0,0 +1,227 @@
+use super::helper_constraints::constrain_lc_with_scalar;
```
mikelodder7: Is there a reason to include this? Is it any more efficient than `set_membership`

libzmix/bulletproofs_amcl/src/r1cs/gadgets/set_non_membership.rs
```diff
@@ -0,0 +1,192 @@
...
+    diff_inv_vars: Vec<AllocatedQuantity>,
+    set: &[FieldElement],
+) -> Result<(), R1CSError> {
+    let set_length = set.len();
+
+    for i in 0..set_length {
+        // Since `diff_vars[i]` is `set[i] - v`, `diff_vars[i]` + `v` should be `set[i]`
+        constrain_lc_with_scalar::<CS>(cs, diff_vars[i].variable + v.variable, &set[i]);
+        // Ensure `set[i] - v` is non-zero
+        is_nonzero_gadget(cs, diff_vars[i], diff_inv_vars[i])?;
```
mikelodder7: This seems to be the only difference with `set_membership`. Can we abstract this?

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+        base_4.push(0);
+    }
+
+    base_4.reverse();
+    base_4
+}
+
+// TODO: ABSTRACT HASH FUNCTION BETTER
+/// Sparse merkle tree with width 4, .i.e each node has 4 children.
+pub struct VanillaSparseMerkleTree_4<'a> {
```
mikelodder7: Could we use a better flavor?!?

libzmix/bulletproofs_amcl/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+authors = ["lovesh harchandani <lovesh.bond@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+rand = "0.6"
+lazy_static = "1.3.0"
+merlin = "1"
+byteorder = "1.3.1"
+
+[dependencies.amcl_wrapper]
```
mikelodder7: Why can't we use libursa/pair/amcl.rs? Let's update it so we don't have to wrap amcl twice.

libzmix/bulletproofs_amcl/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
mikelodder7: We shouldn't need two LICENSE files. This should be a component of zmix and thus be under the same license.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
+use super::helper_constraints::constrain_lc_with_scalar;
```
mikelodder7: Is this done? If so, this uses way more bits than I thought it would. I believed it to be 1 bit/leaf and it looks like 2 bits. Am I wrong?

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+                variable: v,
+                assignment: Some(*i),
+            });
+        }
+    }
+
+    let num_statics = 2;
+    let statics = allocate_statics_for_prover(&mut prover, num_statics);
+
+    let start = Instant::now();
```
mikelodder7: Probably don't want this.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+        sbox_type,
+    )?;
+
+    let total_rounds = hash_params.full_rounds_beginning
+        + hash_params.partial_rounds
+        + hash_params.full_rounds_end;
+    println!("For 4-ary tree of height {} (has 2^{} leaves) and Poseidon rounds {}, no of multipliers is {} and constraints is {}", tree_depth, tree_depth*2, total_rounds, &prover.num_multipliers(), &prover.num_constraints());
+
+    let proof = prover.prove(G, H).unwrap();
+    let end = start.elapsed();
```
mikelodder7: Nor this

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+
+    println!("Proving time is {:?}", end);
```
mikelodder7: Nor this.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+        proof_alloc_scalars.push(AllocatedQuantity {
+            variable: v,
+            assignment: None,
+        });
+    }
+
+    let num_statics = 2;
+    let statics = allocate_statics_for_verifier(&mut verifier, num_statics, g, h);
+
+    let start = Instant::now();
```
mikelodder7: Probably don't want this.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+        leaf_alloc_scalar,
+        leaf_idx_alloc_scalar,
+        proof_alloc_scalars,
+        statics,
+        hash_params,
+        sbox_type,
+    )?;
+
+    verifier.verify(&proof, &g, &h, &G, &H)?;
+    let end = start.elapsed();
```
mikelodder7: Nor this.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
```diff
@@ -0,0 +1,613 @@
...
+
+    println!("Verification time is {:?}", end);
```
mikelodder7: Nor this.

Cargo.toml
lovesh: So that this feature can be developed independently and merged back when more changes are done.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/set_membership_alt.rs
lovesh: These gadgets have not had much peer review. I did not want to discard this one as i took it from another source. The one in `set_membership` is efficient but i have not seen that being used anywhere.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/set_non_membership.rs
lovesh: Thats not the only difference. There is a product in set membership.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
lovesh: Yes, I was trying a better variation but could not get that to work. This is the [one](https://github.com/lovesh/bulletproofs-r1cs-gadgets/blob/master/src/gadget_osmt.rs).

libzmix/bulletproofs_amcl/Cargo.toml
lovesh: Few reasons. Code in amcl.rs is annoying to write/read when you have large number of operations due to lack of operator overloading. Secondly, [amcl_wrapper](https://github.com/lovesh/amcl_rust_wrapper) has abstractions for vectors ops which again makes the code easier to write/read. Apart from the abstractions, amcl_wrapper contains optimizations like Barrett reduction, constant time as well as variable time multi-scalar multiplication and batch inversion. These things do not matter for the stuff we do in libursa now so amcl.rs is sufficient there. But these are necessary for the Bulletproof. So consider the wrapper an external dependency.
Also code in amcl.rs is using a much older version of the amcl and there has been numerous fixes since then. We should get a basic Travis CI setup on miracl/amcl that deploys to crates.io. In the meantime, the wrapper uses crate deployed from this [mirror](https://github.com/lovesh/amcl-mirror).

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
lovesh: The number 4 is for the branching factor of the tree as its for a 4-ary tree.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
lovesh: Ok, will remove.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
lovesh: Same question as above?

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
lovesh: Same question as above?

libzmix/bulletproofs_amcl/src/r1cs/gadgets/vsmt_4.rs
lovesh: Same question as above?

lovesh:
> I don't like the idea of passing tons of parameters to functions. Can we encapsulate this into a struct with a `new` method that takes some of them?

Ok, will try

libzmix/bulletproofs_amcl/LICENSE
lovesh: OK

dcmiddle:
I think the code formatting commit is a result of running `cargo fmt`. If that's the case, a best practice is to run `cargo fmt` before you commit. If you forget and do it later you can always squash the commits together too. Having two separate commits makes the review more complicated.

Also, as always please follow commit message standards: https://chris.beams.io/posts/git-commit/#seven-rules

libzmix/bulletproofs_amcl/LICENSE
dcmiddle: @lovesh can you please clarify the origin of this code?
constraint_system.rs for example appears similar to https://gitlab.dusk.network/kev/bp/blob/7ca35bb2a0ff2d1e7de7aaea7b3faf15ff2856e4/src/r1cs/constraint_system.rs
That code appears under an MIT license. 
If this is the case please contact @dhuseby 
He can explain whether and how to use third party code and licensing obligations.

libzmix/bulletproofs_amcl/README.md
```diff
@@ -0,0 +1,5 @@
+# Bulletproofs
+
```
dcmiddle: What is the intention of this readme?

libzmix/bulletproofs_amcl/README.md
```diff
@@ -0,0 +1,5 @@
...
+1. On a pairing friendly curve. BLS12-381 and BN254 for now. Change value of default feature to change the curve,
```
dcmiddle: Can you please reword this? I think you are conveying that to change that the implementation provides two curve choices that can be selected in cargo.toml.

libzmix/bulletproofs_amcl/README.md
```diff
@@ -0,0 +1,5 @@
...
+2. Using [Apache Milagro](https://github.com/milagro-crypto/amcl).
```
dcmiddle: Is there a reason to call out this dependency vs. others that are probably listed in cargo.toml?

libzmix/bulletproofs_amcl/LICENSE
lovesh: What is meant by origin? Like copy pasted the code as is or following a certain design. The design is copied from (there are some differences though) from dalek's [Bulletproofs implementation here](https://github.com/dalek-cryptography/bulletproofs/blob/develop/src/r1cs/constraint_system.rs). That implementation is MIT.

libzmix/bulletproofs_amcl/README.md
lovesh: To give a summary about this feature.

libzmix/bulletproofs_amcl/README.md
lovesh: Ok

libzmix/bulletproofs_amcl/README.md
lovesh: Because that dependency is the primary differentiator with dalek's implementation. I wanted to answer the question "Why another implementation when there is already one from Dalek."

libzmix/bulletproofs_amcl/LICENSE
dhuseby: So to clarify, this is a fork of the dalek's bp implementation? Do you have any intention to upstream our patches to dalek's code? My recommendation is that we should upstream and then just have a submodule reference to dalek's repo. That adds a bunch of overhead that is a headache but it clarifies a lot of things. IIRC, MIT licensed code is considered "compatible" with HL licensing policies. I need to double check on whether we need a license exception on that code. I don't think so but I want to be certain.

lovesh: This is not a fork but reimplementation (not completely as it is though) of their design using a different math library. Our code is thus incompatible with theirs and we cannot use dalek as a dependency.

libzmix/bulletproofs_amcl/src/utils/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    for _ in 1..n {
+        let _p = &gen * r.last().unwrap();
+        state.extend_from_slice(&_p.to_bytes());
+        r.push(FieldElement::from_msg_hash(&state));
+    }
+    r
+}
+
+#[macro_export]
+macro_rules! vec_to_array {
```
mikelodder7: Can you remove this?

libzmix/bulletproofs_amcl/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "bulletproofs_amcl"
+version = "0.1.0"
+authors = ["lovesh harchandani <lovesh.bond@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+rand = "0.6"
```
mikelodder7: Please update to rand = "0.7"

libzmix/bulletproofs_amcl/src/ipp.rs
```diff
@@ -0,0 +1,386 @@
...
+        let mut a = a_vec.clone();
+        let mut b = b_vec.clone();
+
+        transcript.innerproduct_domain_sep(n as u64);
+
+        let lg_n = n.next_power_of_two().trailing_zeros() as usize;
+        let mut L_vec = G1Vector::with_capacity(lg_n);
+        let mut R_vec = G1Vector::with_capacity(lg_n);
+
+        if n != 1 {
```
mikelodder7: Which section of the paper is this referring? Can you also link the paper?

libzmix/bulletproofs_amcl/src/ipp.rs
```diff
@@ -0,0 +1,386 @@
...
+            L_1.extend(G_R.iter());
+            L_1.extend(H_L.iter());
+            L_1.push(&Q);
+            let mut L_0 = vec![];
+            L_0.extend(a_L.iter());
+            L_0.extend(b_R.iter());
+            L_0.push(&c_L);
+
+            // Inner product of L_1, L_0
+            let L = G1Vector::inner_product_var_time_with_ref_vecs(L_1, L_0).unwrap();
```
mikelodder7: Shouldn't this be the constant time variant?

brentzundel:
I am wondering about the differences between this implementation of bulletproofs and dalek.
Both are in rust, much of the code here seems very similar to the dalek code, or derivative therefrom.
What are the reasons for not just using dalek?

lovesh:
@brentzundel Dalek's implementation uses the Ristretto curve. This one uses BLS12-381 by default but changing to other curves supported by Milagro can be done by changing a feature flag

brentzundel:
Would it make sense to submit a PR to bring the capabilities of this codebase into dalek?

lovesh:
> Would it make sense to submit a PR to bring the capabilities of this codebase into dalek?

The difference in the functionality of the 2 codebases comes from gadgets. Don't know if they would be interested as the gadgets don't add any functionality to the framework but are applications of the framework.

libzmix/bulletproofs_amcl/src/r1cs/gadgets/helper_constraints/poseidon.rs
```diff
@@ -0,0 +1,654 @@
+use crate::errors::R1CSError;
```
khovratovich: I have reviewed the code and it conforms to the spec.

lovesh:
Lets put this PR on hold for 2 days as I am trying to integrate Bulletproofs in Zmix

PR #38, opened 2019-06-21
add new version of Anoncred doc

lovesh:


manudrijvers:
@lovesh, could you clarify how this doc relates to ursa? I was under the impression that Anoncreds is indy-specific.

lovesh:
We already have Anoncreds in Ursa [here](https://github.com/hyperledger/ursa/tree/master/libursa/src/cl) under the name CL signatures.

dcmiddle:
My understanding is that anoncreds is currently part of libursa but we intend to move it to libzmix (both of course still under the same ursa repo). 
I assume that this is an update to the AnonCreds 1.0 spec, updated to reflect https://github.com/hyperledger/ursa/pull/31/commits? But I'm not sure. Especially because this is a PDF update we cannot diff, can you (@lovesh) please update the commit message?

Going forward the plan is to move all spec content into its own repo, [ursa-docs](https://github.com/hyperledger/ursa-docs). That will include the source latex so we can actually track diffs. Please see:
https://lists.hyperledger.org/g/ursa/topic/32056807#148
Once the latex are in the new home, then this content should be removed from the ursa repo.

@manudrijvers if anoncreds in zmix is inconsistent with zklang plans for zmix, please reply to the above thread with your thoughts.  So far I've only seen @hartm respond.

lovesh:
@dcmiddle The changes for the doc do not correspond to that pull request. The primary change in that has not been implemented yet. I will update the commit message but the  pdf contains a Changelog at the bottom and i did update that for this change. 
I will move the latex to ursa-docs. (Btw, I don't see why docs and other specs can't go in the same repo)

manudrijvers:
> @manudrijvers if anoncreds in zmix is inconsistent with zklang plans for zmix, please reply to the above thread with your thoughts. So far I've only seen @hartm respond.

I do actually think it's somewhat incompatible, as we'd like that to build on top of zmix, so express their zero knowledge statements in zklang. On top of that, it seems purely indy-specific. I think this library should provide primitives that are useful to multiple projects.

lovesh:
@manudrijvers The reason this codebase lives in Ursa is to allow Indy to use Ursa. Now, we could have broken down this codebase into modules to do CL signature, several kinds of PoK statements, etc but we don't foresee significant development happening on this codebase due to the move to BBS+ signatures and hence did not spend much resources modularizing it. The next version of Anoncreds will have a more modular structure.

PR #41, opened 2019-07-15
Use multi-pairing during signature verification

lovesh:


mikelodder7:
Looks good I'm doing the same thing in my BBS signature scheme.

PR #42, opened 2019-07-18
Implement curve logic for BLS12-381

mikelodder7:
This is the first of a set of PRs to implement BBS signatures. I have broken it up to help with PR review instead of submitting one big PR.


libzmix/src/utils/bls12_381/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+#[cfg(test)]
+mod tests {
+    use super::{hash_mod_order, random_mod_order};
+    use amcl::bls381::big::BIG;
+    use amcl::bls381::rom;
+    use rand::prelude::*;
+    use rand_chacha::ChaChaRng;
+    use rand_core::SeedableRng;
+
+    #[test]
```
brentzundel: As above, I'd like to see these tests be broken up into a number of different ones, and negative tests added. 
Should there be a test to ensure that the output size is 48 bytes, or that the salt is not less than 32 bytes?

libzmix/src/utils/bls12_381/fr.rs
```diff
@@ -0,0 +1,211 @@
...
+        assert_eq!(FieldOrderElement::from("01"), FieldOrderElement::one());
+        let fr = FieldOrderElement::new();
+        assert_eq!(fr, fr.clone());
+        let salt = b"oiuhgruhqewriuh13k451938475oergnqeiruytkj1b34t098h978123jhk625gv";
+        let domain_sep = b"1234rtgfcdewdfghjnbgt56rtyu3io09iurgiokdjfuighujwkemnj5uyiwhjekr";
+        let foe = FieldOrderElement::from_hash(fr.to_bytes().as_slice(), salt, domain_sep);
+        assert_ne!(foe, fr);
+    }
+
+    #[test]
```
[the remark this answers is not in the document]
mikelodder7: I don't think that's necessary to make each assert its own test. I don't mind making each operator with one positive and one negative.

libzmix/src/utils/bls12_381/mod.rs
mikelodder7: I agree with the test for output size.

libzmix/src/utils/bls12_381/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+    ($name:ident) => {
+        pub fn new() -> Self {
+            $name(random_mod_order::<ThreadRng>(None))
+        }
+    };
+}
+
+/// Generate a random number twice the number of bytes as the curve reduced by mod CURVE_ORDER
+/// Here 96 bytes are randomly generated mod CURVE_ORDER. The result is 48 bytes.
+fn random_mod_order<R: RngCore>(r: Option<&mut R>) -> BIG {
```
lovesh: Why is it generating twice the amount of entropy needed (`compute_big`)?

libzmix/src/utils/bls12_381/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+            fn from(data: &str) -> $name {
+                $name($internal::from_hex(data.to_string()))
+            }
+        }
+
+        impl From<&[u8]> for $name {
+            fn from(data: &[u8]) -> $name {
+                let mut vec = data.to_vec();
+                if data.len() > Self::BYTES_REPR_SIZE {
+                    vec = data[0..Self::BYTES_REPR_SIZE].to_vec();
```
lovesh: I find it better to be explicit and fail if there are more bytes than needed than silently using a truncated version. Caller might find absurd giving 2 different byte slices which are same in first `BYTES_REPR_SIZE` bytes giving the same field element.

lovesh:
@mikelodder7 @dhuseby @dcmiddle @hartm Do you think it would be good to include this AMCL Rust wrapper in Ursa https://github.com/lovesh/amcl_rust_wrapper? It has better abstraction and some more algorithms.

libzmix/src/utils/bls12_381/mod.rs
mikelodder7: To make sure there is no bias is the randomization. Recommendation from @hartm

libzmix/src/utils/bls12_381/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+    static ref GROUP_ORDER: BIG = BIG::new_ints(&rom::CURVE_ORDER);
+}
+
+macro_rules! serialize_impl {
+    ($name:ident, $internal:ident, $visitor:ident) => {
+        impl std::str::FromStr for $name {
+            type Err = std::num::ParseIntError;
+
+            fn from_str(s: &str) -> Result<Self, Self::Err> {
+                Ok($name($internal::from_hex(s.to_string())))
```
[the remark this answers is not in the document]
mikelodder7: Lovesh, your implementation does handle the case for non-hex input but it is not constant time. See https://github.com/lovesh/amcl_rust_wrapper/blob/master/src/field_elem.rs#L436. It bails early. There is also the if with a break. CS does not have any of these.

libzmix/src/utils/bls12_381/mod.rs
mikelodder7: Not opposed to this suggestion.

libzmix/src/utils/bls12_381/mod.rs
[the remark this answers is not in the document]
lovesh: It's not constant time only if the input is invalid or if the input has a smaller size than `MODBYTES*2` where it spends time padding the input. The `break` would always be executed during the last iteration of the loop to avoid an extra shift. And there are always `MODBYTES*2` iterations of the loop and `MODBYTES*2-1` 4-bit left shifts. So it is constant time for valid input and `MODBYTES*2` input of size `MODBYTES*2`

libzmix/src/utils/bls12_381/mod.rs
lovesh: Can you or @hartm provide a reference for this kind of extractor? And do we need a randomness extractor. I thought the entropy source on the machine (assuming you are using a standard machine) should remove any statistical bias.

libzmix/src/utils/bls12_381/mod.rs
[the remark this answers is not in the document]
mikelodder7: That’s still not CS. However you only need constant time to protect secret values like when computing commitments or signing where a private key is used. Serialization I think of any kind doesn’t need to be constant time.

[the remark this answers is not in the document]
lovesh: It's constant time for the cases that matter. Serialization/Deserialization should be constant time. The attacker can learn about private key when you are loading a stored private key or serializing a private key for storage.

libzmix/src/utils/bls12_381/mod.rs
mikelodder7: The Curve order is not exactly the maximum value for 2^381 right. So if you extract 384 bytes you don't get exactly an even distribution and hence the bias towards the lower numbers.

PR #47, opened 2019-08-13
Deb build

mikelodder7:
This PR enables ursa to build a debian package that can be installed on a Debian based OS

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,13 @@ include:
     libursa.dll (Windows)
     libursa.lib (Windows)
 
+
+## Packaging
+Libursa can be packaged for debian builds using [cargo-deb](https://github.com/mmstick/cargo-deb).
+To create a debian package, run the following command with cargo-deb installed
+
+`cargo deb -p ursa` 
+
 ### Libursa Documentation
 
 Libursa API documentation is now available as rust doc in code. See:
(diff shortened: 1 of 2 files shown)


dcmiddle:
```cargo-deb: Asset file path does not match any files: /Users/dcmiddle/project/ursa/libursa/../target/release/libursa.so```

Are you running this in an ubuntu container or right at your local command line?
```warning: This command is for Linux only, and will not make sense when run on other systems```

mikelodder7:
Yes dan I am running in docker Ubuntu but it does work for me if I add a --target x86_64-unknown-linux-gnux32 it works but I had to install a cross compiler. Cargo deb doesn’t distinguish between the current system and what it actually builds. If you run it on OSX it compiles ursa for OSX and this the .so is not what is built. I can add your comments to the README if you think that helps

dcmiddle:
> Yes dan I am running in docker Ubuntu but it does work for me if I add a --target x86_64-unknown-linux-gnux32 it works but I had to install a cross compiler. Cargo deb doesn’t distinguish between the current system and what it actually builds. If you run it on OSX it compiles ursa for OSX and this the .so is not what is built. I can add your comments to the README if you think that helps

Yeah that's probably helpful. Something like "on a debian OS run ...".
Is this something we should add to the travis script as well?

mikelodder7:
> Yeah that's probably helpful. Something like "on a debian OS run ...".
> Is this something we should add to the travis script as well?

I thought about that but didn't think it really adds anything of value to the pipeline because it just builds Ursa then packages it. We are already testing the build portion and the packaging would just be testing the Cargo.toml stuff which should change much other than adding C include files. So I view this as mostly a redundant test if added to the CI.

PR #55, opened 2019-09-11
BBS+ Signatures

mikelodder7:
This adds the BBS+ signature scheme to zmix. The reasonings and math for this can be found [here](https://github.com/hyperledger/ursa-docs/tree/master/specs/anoncreds2)

It has been coded to as close to the same interface that PS signatures. Another PR would be good to create an interface for these two signatures.


libzmix/src/errors.rs
```diff
@@ -0,0 +1,149 @@
...
+use failure::{Backtrace, Context, Fail};
+
+pub mod prelude {
+    pub use super::{BBSError, BBSErrorExt, BBSErrorKind, PSError};
+}
+
+#[derive(Debug, Fail)]
+pub enum PSError {
+    #[fail(
+        display = "Verkey has unequal number of Y and Y_tilde elements. Y={} and Y_tilde={}",
```
brentzundel: This error may mean something to those intimately familiar with the PS paper, but it doesn't convey much to the average user.

libzmix/src/hash_functions/bls12_381_hash.rs
```diff
@@ -176,6 +176,7 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
brentzundel: What is the plan for these ignored tests?

libzmix/src/signatures/bbs/keys.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn key_generate() {
+        let (public_key, _) = generate(0);
+        let bytes = public_key.to_bytes();
+        assert_eq!(bytes.len(), GroupG1_SIZE + 4 + GroupG2_SIZE);
```
brentzundel: A comment here indicating why this assertion matters would be valuable.

libzmix/src/signatures/bbs/keys.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        let (public_key, _) = generate(5);
+        assert_eq!(public_key.message_count(), 5);
+        assert!(public_key.validate().is_ok());
+        let bytes = public_key.to_bytes();
+        assert_eq!(bytes.len(), GroupG1_SIZE * 6 + 4 + GroupG2_SIZE);
```
brentzundel: A comment here indicating why this assertion matters would be valuable.

libzmix/src/signatures/bbs/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod keys;
+pub mod pok_sig;
+pub mod signature;
+
+// TODO: Add "setup" to generate g1 and g2 and use them in-place of G1::generator() and G2::generator()
```
brentzundel: Is there a ticket somewhere describing the work that needs to be done for this TODO?

libzmix/src/signatures/bbs/keys.rs
```diff
@@ -0,0 +1,106 @@
...
+        } else {
+            Ok(())
+        }
+    }
+}
+
+pub fn generate(message_count: usize) -> (PublicKey, SecretKey) {
+    let secret = FieldElement::random();
+
+    // XXX: Choosing G2::generator() temporarily. The generator should be a setup parameter in practice
```
brentzundel: Is there a ticket somewhere describing the work that needs to be done for this?

libzmix/src/signatures/bbs/signature.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    #[test]
+    fn signature_serialization() {
+        let sig = Signature {
+            a: G1::random(),
+            e: FieldElement::random(),
+            s: FieldElement::random(),
+        };
+        let bytes = sig.to_bytes();
+        assert_eq!(bytes.len(), GroupG1_SIZE + MODBYTES * 2);
```
brentzundel: Once again, a description of with this assertion matters would be helpful for those who haven't read the paper.

libzmix/src/errors.rs
```diff
@@ -0,0 +1,149 @@
+use commitments::pok_vc::PoKVCError;
+use failure::{Backtrace, Context, Fail};
+
+pub mod prelude {
+    pub use super::{BBSError, BBSErrorExt, BBSErrorKind, PSError};
+}
+
+#[derive(Debug, Fail)]
+pub enum PSError {
```
lovesh: This error is already defined in `ps`

libzmix/src/signatures/bbs/keys.rs
```diff
@@ -0,0 +1,106 @@
...
+    let w = &G2::generator() * &secret;
+    let mut h = Vec::new();
```
lovesh: `h0` and all `h` must be NUMS so hash some publicly known string(s) to generate them.

libzmix/src/signatures/bbs/keys.rs
```diff
@@ -0,0 +1,106 @@
...
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn key_generate() {
+        let (public_key, _) = generate(0);
```
lovesh: Nitpick: Does it make sense to allow 0 as an arg to `generate`?

libzmix/src/signatures/bbs/signature.rs
```diff
@@ -0,0 +1,281 @@
...
+        verkey: &PublicKey,
+    ) -> Result<Self, BBSError> {
+        check_verkey_message(messages, verkey)?;
+        let e = FieldElement::random();
+        let s = FieldElement::random();
+        let b = compute_b_const_time(&G1::new(), verkey, messages, &s, 0);
+        let mut exp = signkey.clone();
+        exp += &e;
+        exp.inverse_mut();
+        let a = b * exp;
```
lovesh: Very unlikely that `a` would be the identity element but maybe we should still check that it's not since the verifies does not accept proof where `a` is identity.

libzmix/src/signatures/bbs/signature.rs
```diff
@@ -0,0 +1,281 @@
...
+    ) -> Result<Self, BBSError> {
+        if messages.len() >= verkey.message_count() {
+            return Err(BBSError::from_kind(
+                BBSErrorKind::SigningErrorMessageCountMismatch(
+                    verkey.message_count(),
+                    messages.len(),
+                ),
+            ));
+        }
+        let e = FieldElement::random();
```
lovesh: Another nitpick: Some code duplication can be avoided by using common code from `new`.

libzmix/src/signatures/bbs/signature.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut exp = signkey.clone();
+        exp += &e;
+        exp.inverse_mut();
+        let a = b * exp;
+        Ok(Signature { a, e, s })
+    }
+
+    // Once signature on committed attributes (blind signature) is received, the signature needs to be unblinded.
+    // Takes the blinding used in the commitment.
+    pub fn get_unblinded_signature(&self, blinding: &FieldElement) -> Self {
```
lovesh: Where is this function used?

libzmix/src/signatures/bbs/mod.rs
lovesh: Some issuers might not trust the generation of `G1::generator()` and `G2::generator()`. The verifier should be convinced that the discrete log them wrt. others is not known.

libzmix/src/signatures/bbs/keys.rs
lovesh: Same as above

libzmix/src/signatures/bbs/keys.rs
lovesh: How would you convince someone that you generated them randomly?

hartm:
> It has been coded to as close to the same interface that PS signatures. Another PR would be good to create an interface for these two signatures.

I can help with this.  How much of the interface do you want to expose, and what functionalities do you want to integrate?

mikelodder7:
> I can help with this. How much of the interface do you want to expose, and what functionalities do you want to integrate?

The key gen, signing, verify and proof of signature is probably the best place to start

libzmix/src/signatures/bbs/keys.rs
lovesh: I was wrong about `h`s. Since they are specific to one issuer, it's fine to construct them this way

PR #27, opened 2019-05-14
Travis CI pipeline

mikelodder7:
This enables the *travis-ci* pipeline for OSX and Linux.  See [here](https://travis-ci.org/mikelodder7/ursa/builds/532517338) for how it works.

I added Cargo.lock to be versioned per recommendation by @dhuseby who said
> The idea of checking in the cargo.lock file is for reproducible/deterministic builds. We want to track the contents of that over time so that we know exactly what dependencies were used to support specific builds of Ursa.

The rest of the changes are just *rustfmt* and *cargo clippy* fixes.

PR #61, opened 2019-09-22
remove libursa Cargo.lock

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

I think that, in general, Cargo.lock should not be in version control of libraries like ursa.

libindy uses ursa but ursa is using an older version of openssl-sys than libindy

(Kind of hoping that using the same, newer versions help with libindy's CI-run failures on Ubuntu 18.04)


JAG-UK:
I can't help thinking this should go through an RFC before we just do it.  I really don't like the way Cargo dependencies are handled in composite projects and I find the whole thing with libraries importing different copies of deps a little bit dangerous but at the same time having libnrsa lose control of which versions of underlying crypto libraries it uses is instinctively a bad move for me.  

2 concrete thoughts:
 - The main purpose of Ursa is to make it easy for people to access safe crypto and not have to think too hard for themselves or risk putting themselves at the mercy of their own mistakes.  This includes things like picking up the correct OpenSSL versions etc. And don't let's get started on the possibility of Forks.
 - If we ever wanted a FIPS version of Ursa we'd have to pin the dependencies and by definition all copies/alternative versions brought in by the application would be outside the boundary.  Even without a strict FIPS requirement it's a pretty basic requirement of any security library to have reproducible builds, which we lose in this case: change control/approved configurations then migrate to the relying application which is impossible for Ursa to handle.

I'm not saying I object, but I think this is an important philosophical discussion that deeply affects what Ursa does/can promise, and I don't remember having discussed it in the calls before.

vaporos:
This may solve similar issues with libsecp256 versions we have had trying to use ursa.

AxelNennker:
Maybe we can care about binary build reproducibility when it is needed? And then re-introduce Cargo.lock.

I see this problem: if libursa uses e.g. openssl 1.0.2g-1ubuntu4.15 on Ubuntu 16.04 and libindy is using openssl 1.1.x on Ubuntu 18.04 then this might lead to link errors because openssl methods got removed and renamed between openssl 1.0.x and 1.1.x

AxelNennker:
Could you please update Cargo.lock to point to the latest and the same openssl-sys as in libindy that is 0.9.49, please?

These two lines should both point to 0.9.49
https://github.com/hyperledger/ursa/blob/master/libursa/Cargo.lock#L521
https://github.com/hyperledger/indy-sdk/blob/master/libindy/Cargo.lock#L908

Regarding reproduceable builds: Maybe add a task to update Cargo.lock each Tuesday? :-)
Or whatever the teams workload. The Rust ecosystem is rapidly evolving and staying behind will hurt again and again.

AxelNennker:
Does this PR actually remove the Cargo.lock file. There is the Cargo.lock in the root of ursa and that should lock libursa as well doesn't it. So deleting libursa/Cargo.lock does not really change anything for libursa, right?
Maybe I need some education on cargo lock files.

mikelodder7:
Removing libursa/Cargo.lock should be fine. Rust is only using the workspace level one

vaporos:
The project I'm working on that was using Ursa has removed support from Ursa temporarily over a related build issue, pending our ability to resolve the issue. The FAQ Dan linked in the other PR clearly says best practice is to not check-in Cargo.lock for library projects for the reasons we experienced with build issues. AFAIK, there is no way to "opt out" of the Cargo.lock if you are consuming the library.

AxelNennker:
So this PR can be merged after the "merger" resolved the merge conflicts?
That is another problem with Cargo.lock files. merge does not know how to handle the format.

AxelNennker:
When will the new version of ursa with [openssl-sys 0.9.49](https://crates.io/search?q=openssl-sys) be published to https://crates.io/search?q=ursa

dcmiddle:
We may want to revisit the decision to source control Cargo.lock. 
I did not search exhaustively but I did look at similar projects like all the RustCrypto repos and do not see them tracking Cargo.lock
https://github.com/search?q=org%3ARustCrypto+Cargo.lock&unscoped_q=Cargo.lock

Per discussion in the earlier [PR](https://github.com/hyperledger/ursa/pull/27) The cargo reference says not to track Cargo.lock for libraries:
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

Regarding comment in that earlier PR that it's needed for running cargo audit, there are [new instructions](https://github.com/RustSec/cargo-audit/blame/master/README.md#L37) in the cargo audit readme how to work with travis.

Namely:
```
before_script:
  - cargo install --force cargo-audit
  - cargo generate-lockfile```

mikelodder7:
I want to hear @dhuseby's feedback on this. He was the original push for including Cargo.lock.

AxelNennker:
I would kindly ask that for now the updated Cargo.lock is published to crates.io
Could somebody with publish rights please do that?
That would be https://crates.io/crates/ursa/0.2.0-dev-2 then, I guess.

mikelodder7:
@AxelNennker https://crates.io/crates/ursa/0.2.0-dev-2

mikelodder7:
> We may want to revisit the decision to source control Cargo.lock.
> I did not search exhaustively but I did look at similar projects like all the RustCrypto repos and do not see them tracking Cargo.lock
> https://github.com/search?q=org%3ARustCrypto+Cargo.lock&unscoped_q=Cargo.lock
> 
> Per discussion in the earlier [PR](https://github.com/hyperledger/ursa/pull/27) The cargo reference says not to track Cargo.lock for libraries:
> https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
> 
> Regarding comment in that earlier PR that it's needed for running cargo audit, there are [new instructions](https://github.com/RustSec/cargo-audit/blame/master/README.md#L37) in the cargo audit readme how to work with travis.
> 
> Namely:
> 
> ```
> before_script:
>   - cargo install --force cargo-audit
>   - cargo generate-lockfile
> ```

Yes but the intention with those libraries is to be solely Rust crates. Ursa is meant for more than that. It includes FFI and is meant to be a dependency for multiple Hyperledger Projects.

However, The Rust team wrote this too: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

jovfer:
@AxelNennker @mikelodder7 
Based on my experience with Indy projects I'd like to recommend the following
1) exclude Cargo.lock from crates.io artifact at all
2) keep Cargo.lock in the repo and use it to have reproducible artifact of Ursa in a form of system library.

> Yes but the intention with those libraries is to be solely Rust crates. Ursa is meant for more than that. It includes FFI and is meant to be a dependency for multiple Hyperledger Projects.

> However, The Rust team wrote this too: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries.

I think I have similar position with Mike...
We have 2 completely different (from cargo point of view) artifacts of Ursa:
* cargo crate - it SHOULD NOT contains Cargo.lock as recommended at the link above (because it's consumed in a form of rust/cargo dependency)
* system libraries (dynamic and/or static) - from cargo point of view it's more like final binary, so it should be based on Cargo.lock tracked by VCS (it's consumed as C-callable artifact, not rust crate)

mikelodder7:
This has been resolved with other PRs. Going to close

PR #65, opened 2019-09-30
Fix secp256k1 test

dcmiddle:
Fix secp256k1 test and add it to CI so that we can detect if it breaks again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,4 +38,6 @@ script:
   - cargo build
   - cargo build --no-default-features --features=portable
   - cargo test
+  - cargo run --manifest-path ./libursa/Cargo.toml --bin test_secp256k1 --features="benchmarksecp256k1"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -38,4 +38,6 @@ script:
   - cargo build
   - cargo build --no-default-features --features=portable
   - cargo test
+  - cargo run --bin test_secp256k1 --features="benchmarksecp256k1"
+  - cargo run --bin test_ed25519 --features="benchmarked25519"
```
lovesh: Why are these tests in `libursa/bin` and not somewhere in `libursa/src/signatures/`?

mikelodder7: These are binaries that run and are not libraries.

lovesh: Which binaries are being tested? The test appears to be calling library functions

mikelodder7: It's testing the speed of secp256k1 signatures and ed25519 signatures respectively. These tests can be run against portable or default mode.

PR #66, opened 2019-10-02
Added crates.io packaging script

mikelodder7:
Added a script that prepares libursa to be published to crates.io.

Also updated the **authors** to Hyperledger Ursa Contributors.

### DIFF
--- a/libursa/Cargo.toml
+++ b/libursa/Cargo.toml
@@ -1,13 +1,28 @@
 [package]
 name = "ursa"
 version = "0.2.0"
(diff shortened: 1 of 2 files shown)


libursa/Cargo.toml
```diff
@@ -1,13 +1,28 @@
 [package]
 name = "ursa"
 version = "0.2.0"
-authors = ["Michael Lodder <redmike7@gmail.com>"]
```
mac-arrap: Thank you for not taking all the credit!

PR #71, opened 2019-10-14
Implement BLS signatures on BLS12-381

mikelodder7:
This provides two BLS implementations on the BLS12-381 curve. Normal and Small. 
Normal = Signatures on G2, Public Keys on G1
Small = Signatures on G1, Public Keys on G2.

I added benchmarks and tests.


libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+    bls_tests_impl!();
+}
+
+/// This version is the small BLS signature scheme
+/// with the public key group in G2 and signature group in G1.
+/// 97 bytes signatures and 192 byte public keys
+///
+/// This results in smaller signatures but slower operations and bigger public key.
+/// This is good for situations where space is a consideration and verification is infrequent
+pub mod small {
```
lovesh: Rather than duplicating the same logic as present in `mod normal` in here, you can use macros if you want to use both configurations in the same compiled code. Or use features that put signature in G1 or G2 if you want the configuration to be picked at compile time

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+    use super::*;
```
lovesh: My comments from `mod normal` apply here as well as the code here is same as in `mod normal`

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+    group_elem::GroupElement,
+    group_elem_g1::G1,
+    group_elem_g2::G2,
+    extension_field_gt::GT,
+    constants::{MODBYTES, GroupG1_SIZE},
+    types_g2::GroupG2_SIZE
+};
+use CryptoError;
+
+pub const PRIVATE_KEY_SIZE: usize = MODBYTES;
```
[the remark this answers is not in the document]
mikelodder7: Primarily there to be consistent with the other signature schemes.

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct PublicKey(G1);
+
+    impl PublicKey {
+        pub fn new(sk: &PrivateKey) -> Self {
+            sk.into()
+        }
+
+        // Create an aggregate public key without rogue mitigation
+        pub fn aggregate(&mut self, pks: &[PublicKey]) {
```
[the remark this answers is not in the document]
mikelodder7: Should it be `combine`?

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+        }
+
+        pub fn from_bytes(bytes: &[u8]) -> Result<Self, CryptoError> {
+            Ok(PublicKey(G1::from_bytes(bytes).map_err(|e| CryptoError::ParseError(format!("{:?}", e)))?))
+        }
+    }
+
+    impl From<&PrivateKey> for PublicKey {
+        fn from(sk: &PrivateKey) -> Self {
+            PublicKey(&G1::generator() * sk)
```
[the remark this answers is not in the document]
mikelodder7: But we are tied to the AMCL implementation. If we switch to a different provider, most of the code will change.

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+        fn from(keys: &[PublicKey]) -> Self {
+            // To combat the rogue key attack,
+            // compute (t_1,…,t_n)←H1(pk_1,…,pk_n) ∈ R_n
+            // output the aggregated public key
+            // as described in section 3.1 from https://eprint.iacr.org/2018/483
+            let mut bytes = Vec::new();
+            for k in keys {
+                bytes.extend_from_slice(k.0.to_bytes().as_slice());
+            }
+            AggregatedPublicKey(keys.iter().fold(G1::identity(), |apk, k| {
```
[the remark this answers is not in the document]
mikelodder7: So does the code become?
```
apk + k.scalar_mul_const_time(&FieldElement::from_msg_hash(h.as_slice())
```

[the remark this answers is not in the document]
mikelodder7: Its already calling that in the Multiply operator implementations.

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+    /// versus verifying them one by one as it reduces the number of computed pairings
+        pub fn batch_verify(inputs: &[(&[u8] /* message */, &AggregatedSignature, &AggregatedPublicKey)]) -> bool {
+            // To combat the rogue key attack and avoid checking for distinct messages
+            // use batch verification as described in the end of section 3.1 from https://eprint.iacr.org/2018/483
+            let mut pairs = Vec::new();
+            let mut sig = G2::identity();
+            for (msg, asg, apk) in inputs {
+                let random_exponent = FieldElement::random();
+                let g2 = G2::from_msg_hash(msg);
+                sig += &asg.0 * &random_exponent;
```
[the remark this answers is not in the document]
mikelodder7: Its already calling that in the Multiply operator implementations.

libursa/src/signatures/bls.rs
mikelodder7: Except that the math is slightly different in the verify steps. I've tried to make many things common with macros and ultimately didn't work out well and made the code more confusing.

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
lovesh: Where do you use this method?

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
lovesh: I mean if we wanted another BLS implementations (which use the same hash technique) to verify our signatures or verify signatures from other implementations.

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
lovesh: I meant you could hold each `h` and `k.0` in the fold in vectors of field and group elements respectively and then do a multi-scalar multiplication once

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,693 @@
...
+            let ate_pairs = pairs.iter().map(|(g1, g2)| (g1, g2)).collect();
+            GT::ate_multi_pairing(ate_pairs).is_one()
+        }
+
+        pub fn batch_verify(inputs: &[(&[u8], &Signature, &PublicKey)]) -> bool {
+            // To avoid rogue key attacks, you must use proof of possession or `AggregateSignature::batch_verify`
+            // This function just avoids checking for distinct messages and
+            // uses batch verification as described in the end of section 3.1 from https://eprint.iacr.org/2018/483
+            let mut pairs = Vec::new();
+            let mut sig = G2::identity();
```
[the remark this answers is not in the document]
lovesh: How is it calling? I meant you can compute sig by doing multi-scalar multiplication between vectors of randomness (`random_exponent` in the loop below) and `apk.0`

libursa/src/signatures/bls.rs
lovesh: The math should not be different. Can you point out the specific instance?

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
mikelodder7: https://github.com/mikelodder7/ursa/blob/bls/libursa/src/signatures/bls.rs#L80

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
mikelodder7: How do we pass these parameters without gunking up the interface?

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
mikelodder7: Can you write an example?

libursa/src/signatures/bls.rs
mikelodder7: https://github.com/mikelodder7/ursa/blob/bls/libursa/src/signatures/bls.rs#L372
https://github.com/mikelodder7/ursa/blob/bls/libursa/src/signatures/bls.rs#L618

They have to be different because the pairing expects the values to be G1, G2. Just using a macro to change G1->G2 and G2->G1 doesn't work there because you have to swap the variables too.

mikelodder7: The math isn't necessarily different. Its just an ordering of the variables.

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
lovesh: `combine` is fine, `add` is fine, i don't care much about the name unless it doesn't collide with an existing

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
lovesh: You can wrap the params in the struct and pass the struct as a single arg (there is one param now anyway), like [this](https://github.com/lovesh/signature-schemes/tree/master/bls#generate-keys)

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
lovesh: https://github.com/lovesh/signature-schemes/blob/master/bls/src/multi_sig_slow.rs#L51

libursa/src/signatures/bls.rs
lovesh: The macro takes a function as one of the arguments which does the pairing. One function's signature will be something like `Fn(&G1, &G2, &G1, &G2)` and the other's will be `Fn(&G2, &G1, &G2, &G1)` but both call the pairing function. Or you can keep the signature same but change the order of params in the implementation.

libursa/src/signatures/bls.rs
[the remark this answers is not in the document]
mikelodder7: Well the default curve generator works so I'm seeing a need to change it.

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,706 @@
...
+            Signature(&SignatureGroup::from_msg_hash(message) * sk)
+        }
+
+        pub fn new_with_rk_mitigation(
+            message: &[u8],
+            sk: &PrivateKey,
+            pk_index: usize,
+            pks: &[PublicKey],
+        ) -> Self {
+            // To combat the rogue key attack,
```
lovesh: The comment seems irrelevant. Where is `(t_1,…,t_n)←H1(pk_1,…,pk_n) ∈ R_n` happening?

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,706 @@
...
```
lovesh: Same comment as above

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,706 @@
...
+                for k in keys {
+                    bytes.extend_from_slice(k.to_bytes().as_slice());
+                }
+                AggregatedPublicKey(keys.iter().fold(Generator::identity(), |apk, k| {
+                    // The position of the ith public key in the byte array
+                    // of the hash doesn't matter as much as its included twice.
+                    // For convenience, its appended to the end
+                    let mut h = bytes.clone();
+                    h.extend_from_slice(k.0.to_bytes().as_slice());
+                    apk + &k.0 * &FieldElement::from_msg_hash(h.as_slice())
```
lovesh: Domain separation?

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,706 @@
...
+            // compute (t_1,…,t_n)←H1(pk_1,…,pk_n) ∈ R_n
+            // output the aggregated public key
+            // as described in section 3.1 from https://eprint.iacr.org/2018/483
+            let mut bytes = Vec::new();
+            for k in pks {
+                bytes.extend_from_slice(k.to_bytes().as_slice());
+            }
+            bytes.extend_from_slice(pks[pk_index].to_bytes().as_slice());
+            let a = FieldElement::from_msg_hash(bytes.as_slice());
+            Signature(SignatureGroup::from_msg_hash(message) * sk * &a)
```
lovesh: Domain separation?

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,706 @@
...
+        // handles this using other techniques like proof of possession
+        pub fn combine(&mut self, signatures: &[Signature]) {
+            for sig in signatures {
+                self.0 += &sig.0;
+            }
+        }
+
+        // Verify a signature generated by `new`
+        pub fn verify(&self, message: &[u8], pk: &PublicKey, g: &Generator) -> bool {
+            let hash = SignatureGroup::from_msg_hash(message);
```
lovesh: Domain separation?

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,706 @@
...
+
+        // Caller should aggregate all signatures into `self` by using `combine`.
+        // Messages must be distinct
+        // `inputs` is a slice of message - public key tuples
+        // Multisignature verification
+        pub fn verify_multi(&self, inputs: &[(&[u8], &PublicKey)], g: &Generator) -> bool {
+            let mut msg_check = ::std::collections::HashSet::new();
+            let mut pairs = Vec::new();
+            for (msg, pk) in inputs {
+                let hash = SignatureGroup::from_msg_hash(&msg);
```
lovesh: Domain separation?

libursa/src/signatures/bls.rs
mikelodder7: I don’t think it’s necessary just like for simple signatures.

lovesh: For what kind of signatures do you think it's necessary? FYI, the paper says that you should use different hash functions for hashing verkeys and messages (H1 and H0 respectively)

mikelodder7: It's up to the caller to determine the domains to use. The caller can prepend or append the domain separation to the message that's being signed. That's no different than Ed25519, Secp256k1, or RSA.

libursa/src/signatures/bls.rs
mikelodder7: in the code right below it.

libursa/src/signatures/bls.rs
mikelodder7: Same response as above

libursa/src/signatures/bls.rs
lovesh: I don't see how. Maybe what you're doing works as well but it does not follow the comment. `(t_1,…,t_n)←H1(pk_1,…,pk_n)` means there is one hash function (`H1`) call to generate `n` field elements, `t_1,t_2,..t_n`. One way to construct `H1` would be to use an XOF that takes `pk_1||…||pk_n` as input and outputs result of size `n*sizeof(t_i)`. This output is then used to construct `t_1,t_2,..t_n`.

libursa/src/signatures/bls.rs
lovesh: The domains are determined, one is the message's and the other is verkey's. Why does the caller cares whether you simulate different random oracles using different hash functions or using different prefix with the same hash function?

libursa/src/signatures/bls.rs
mikelodder7: @lovesh, The caller is the one that knows what is being signed unless we have a function for each specific purpose. The general practice is to prepend or append bytes to the message being signed. If we want to be able to select a different XOF depending on the type of data being signed we should expose that. But in general, signing libraries do not handle domain separation directly. OpenSSL, other SSL variants, Libsodium, libsecp256k1 all do not. Dalek crypto offers a byte array as an optional parameter [here](https://github.com/dalek-cryptography/ed25519-dalek/blob/5870607aa770f8f08114fb8a8eae7e605b82b7c9/src/secret.rs#L482) but doesn't enforce it. Why?  Because it's up to the caller to know what they are signing. The library just supplies the methods to sign. The library has no idea what is being signed unless it has a specific function called `sign_public_key` or `sign_general_message`. In those cases, the only difference between them is the XOF  to hash to point or a fixed prefix. I believe the caller would like to have these defined for their system parameters. Fundamentally they are exactly the same in every way minus that detail. Having two methods like that is okay but the majority of the time, only the one method is used. So can you give me a good reason why we should add domain separation into the library vs punting to the user to add their own domain separation?

lovesh: How is XOF relevant here? 
**As i said in my previous comment, the separation of domains is done between messages and verkeys.** 
All messages being signed will have the same prefix prepended/appended before hashing them. This prefix will be different than the prefix prepended/appended before hashing public keys but public keys will have the same prefix. The reason for this separation is to get separate random oracles (as the paper demands it and will be needed for security proof) built from the same hash function. 

> So can you give me a good reason why we should add domain separation into the library vs punting to the user to add their own domain separation?

Because a domain separator chosen by the user will not have any special property than a separator chosen by the implementation. It will complicate the interface and then puts the burnden on the caller(s) to use differen separators for messages and verkeys.

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,709 @@
...
+/// instead of wrapping it as a private field
+pub type PrivateKey = FieldElement;
+
+macro_rules! bls_impl {
+    ($pk_size:expr, $sig_size:expr, $pk_group:ident, $sig_group:ident, $ate_2_pairing_is_one:ident, $set_pairs:ident) => {
+        pub const PUBLIC_KEY_SIZE: usize = $pk_size;
+        pub const SIGNATURE_SIZE: usize = $sig_size;
+
+        pub const MESSAGE_CONTEXT: &[u8; 20] = b"for signing messages";
+        pub const PUBLICKEY_CONTEXT: &[u8; 47] = b"for signing public keys for proof of possession";
```
lovesh: There needs to be a separator used while hashing verkeys during aggregation. [Here](https://github.com/lovesh/signature-schemes/blob/master/bls/src/multi_sig_slow.rs#L24) is an example, the `hash_verkeys` uses a domain separator

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,709 @@
...
+                for k in keys {
+                    bytes.extend_from_slice(k.to_bytes().as_slice());
+                }
+                AggregatedPublicKey(keys.iter().fold(Generator::identity(), |apk, k| {
+                    // The position of the ith public key in the byte array
+                    // of the hash doesn't matter as much as its included twice.
+                    // For convenience, its appended to the end
+                    let mut h = bytes.clone();
+                    h.extend_from_slice(k.0.to_bytes().as_slice());
+                    apk + &k.0 * &FieldElement::from_msg_hash(h.as_slice())
```
lovesh: `apk` can computed using multi-scalar multiplication `multi_scalar_mul_var_time`. Have a look [here](https://github.com/lovesh/signature-schemes/blob/master/bls/src/multi_sig_slow.rs#L48)

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,709 @@
...
+            ) -> bool {
+                // To avoid rogue key attacks, you must use proof of possession or `AggregateSignature::batch_verify`
+                // This function just avoids checking for distinct messages and
+                // uses batch verification as described in the end of section 3.1 from https://eprint.iacr.org/2018/483
+                let mut pairs = Vec::new();
+                let mut sig = SignatureGroup::identity();
+                for (msg, asg, apk) in inputs {
+                    let random_exponent = FieldElement::random();
+                    let hash = hash_msg(msg, context);
+                    sig += &asg.0 * &random_exponent;
```
lovesh: `sig` can computed using multi-scalar multiplication `multi_scalar_mul_const_time`

libursa/src/signatures/bls.rs
```diff
@@ -0,0 +1,709 @@
...
+                g: &Generator,
+            ) -> bool {
+                // To combat the rogue key attack and avoid checking for distinct messages
+                // use batch verification as described in the end of section 3.1 from https://eprint.iacr.org/2018/483
+                let mut pairs = Vec::new();
+                let mut sig = SignatureGroup::identity();
+                for (msg, asg, apk) in inputs {
+                    let random_exponent = FieldElement::random();
+                    let hash = hash_msg(msg, context);
+                    sig += &asg.0 * &random_exponent;
```
lovesh: `sig` can computed using multi-scalar multiplication

libursa/src/signatures/bls.rs
mikelodder7: You might want to look at the code you referenced because it doesn’t even do what are suggesting. The DS is good enough as it is

lovesh: I looked before suggesting. `hash_verkeys` is adding the separator `VERKEY_DOMAIN_PREFIX` before hashing.

PR #73, opened 2019-10-19
Update Cargo.toml

AxelNennker:


mikelodder7:
This will break our ed25519 code because ed25519-dalek isn't compatible with rand=0.7 yet. Believe me I've tried.

danintel:
You need to sign your commit with a DCO. This is a Linux Foundation requirement. See https://github.com/hyperledger/ursa/blob/master/CONTRIBUTING.md on how to contribute source code.

AxelNennker:
> This will break our ed25519 code because ed25519-dalek isn't compatible with rand=0.7 yet. Believe me I've tried.

Cool, you fixed the problem at the root.
https://github.com/dalek-cryptography/ed25519-dalek/pull/101

AxelNennker:
amending the patch for dco does not work... I'll close it after @mikelodder7 PR is merged. 
Somebody/I can do the update to rand 0.7 then.

In case you are interested why I am doing this... I am trying to get of multiple versions of the same crate to be compiled into libindy.
See `cargo tree -d`
Which is tedious and might annoy people. The Rust ecosystem is moving fast so every crate author is behind somewhere all the time.

mikelodder7:
Unfortunately, we also need `num-bigint` to update to rand=0.7 before Ursa can.

AxelNennker:
https://github.com/rust-num/num-bigint/pull/110 fixes this

⁣Get BlueMail for Android ​

mikelodder7:
In the ursa call today, we discussed closing this PR. Is there anything else we should discuss before doing so?

AxelNennker:
I am on vacation, so can't check until Sa.

Libindy uses 0.7

In general I find it kind of unacceptable to be stuck at a lower version (for any crate)

⁣Get BlueMail for Android ​

mikelodder7:
I don't like being stuck at a lower version either. However, we can't always jump to the latest. Controlled marches.

mikelodder7:
Understandable, and we do try to be up-to-date as much as possible. However, we have to do so in a controlled way that doesn't break code. We watch the dependencies closely to make sure we can stay up-to-date. Thanks for keeping an eye on this.

PR #75, opened 2019-10-23
Update documentation and docker files

mikelodder7:
The libsodium link was moved for 1.0.18 and the docker files were using 1.0.16. This fixes both of those references to use 1.0.18 and the correct link.

Thanks to @burdettadam for finding the discrepancy


brentzundel:
@danintel https://github.com/jedisct1/libsodium is the official github repo for libsodium, so it looks like we're not moving from a public archive to someone's personal repository, but rather moving from the official public archive to the official github repository.

PR #76, opened 2019-11-05
Added software encryption and cleaned up TOML files

mikelodder7:
This adds some AEAD encryption algorithms to ursa. This is related to Ursa-RFC [6](https://github.com/hyperledger/ursa-rfcs/pull/6).

### DIFF
--- a/libursa/src/encryption/mod.rs
+++ b/libursa/src/encryption/mod.rs
@@ -6,6 +6,7 @@ use rand::{rngs::OsRng, RngCore};
 
 pub mod symm;
 
+// Helpful for generating bytes using the operating system random number generator
 pub fn random_vec(bytes: usize) -> Result<Vec<u8>, Error> {
     let mut value = vec![0u8; bytes];
     let mut rng = OsRng::new().map_err(|_| Error)?;
--- a/libursa/src/encryption/mod.rs
+++ b/libursa/src/encryption/mod.rs
@@ -1,0 +1,20 @@
+//! Encryption is categorized as public key encryption or symmetric encryption
+//! `symm` provides symmetric AEAD cryptographic algorithms.
+use aead::generic_array::{ArrayLength, GenericArray};
+use aead::Error;
+use rand::{rngs::OsRng, RngCore};
+
+pub mod symm;
+
+pub fn random_vec(bytes: usize) -> Result<Vec<u8>, Error> {
+    let mut value = vec![0u8; bytes];
+    let mut rng = OsRng::new().map_err(|_| Error)?;
+    rng.fill_bytes(value.as_mut_slice());
+    Ok(value)
+}
+
+pub fn random_bytes<T: ArrayLength<u8>>() -> Result<GenericArray<u8, T>, Error> {
+    Ok(GenericArray::clone_from_slice(
+        random_vec(T::to_usize())?.as_slice(),
+    ))
+}
--- a/libursa/src/encryption/symm/aescbc.rs
+++ b/libursa/src/encryption/symm/aescbc.rs
@@ -1,0 +1,141 @@
+use super::Encryptor;
+use aead::{
+    generic_array::{
+        typenum::{Unsigned, U0, U16, U32, U48, U64},
+        GenericArray,
+    },
+    Aead, Error, NewAead, Payload,
+};
+use aes::{Aes128, Aes256};
+use block_modes::block_padding::Pkcs7;
+use block_modes::{BlockMode, Cbc};
+use hmac::{Hmac, Mac};
+#[cfg(feature = "serialization")]
+use serde::de::{Deserialize, Deserializer, Error as DError, Visitor};
+#[cfg(feature = "serialization")]
+use serde::ser::{Serialize, Serializer};
+use sha2::{Sha256, Sha512};
+use subtle::ConstantTimeEq;
+use zeroize::Zeroize;
+
+type Aes128Cbc = Cbc<Aes128, Pkcs7>;
+type Aes256Cbc = Cbc<Aes256, Pkcs7>;
+type HmacSha256 = Hmac<Sha256>;
+type HmacSha512 = Hmac<Sha512>;
+
+macro_rules! aes_cbc_hmac_impl {
+    ($name:ident, $algokeysize:ident, $keysize:ident, $noncesize:ident, $tagsize:ident, $algo:ident, $hash:ident, $visitor:ident) => {
+        #[derive(Debug, Clone, Eq, PartialEq)]
+        pub struct $name {
+            key: GenericArray<u8, $keysize>,
+        }
+
+        impl Encryptor for $name {
+            type MinSize = U48;
+        }
+
+        impl NewAead for $name {
+            type KeySize = $keysize;
+
+            fn new(key: GenericArray<u8, $keysize>) -> Self {
+                Self { key }
+            }
+        }
+
+        impl Aead for $name {
+            type NonceSize = $noncesize;
+            type TagSize = $tagsize;
+            type CiphertextOverhead = U0;
+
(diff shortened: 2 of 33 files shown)


lovesh:
Why do we need AEC-CBC HMAC when we are supporting AES-GCM since GCM is faster and can use native instructions as well? Are we trying to address some of GCM's issues as described [here](https://crypto.stackexchange.com/a/18422)? If yes, then which ones?

mikelodder7:
> Why do we need AEC-CBC HMAC when we are supporting AES-GCM since GCM is faster and can use native instructions as well? Are we trying to address some of GCM's issues as described [here](https://crypto.stackexchange.com/a/18422)? If yes, then which ones?

First reason FIPS. Certain levels of FIPS only allow AES-CBC. Second, (Fernet)[https://github.com/fernet/spec/blob/master/Spec.md] is a popular way to create tokens. I'm not trying to avoid the issues you put in the link. AES-CBC can use native instructions too. They both can use the AES-NI.

libursa/src/encryption/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+
+pub fn random_vec(bytes: usize) -> Result<Vec<u8>, Error> {
+    let mut value = vec![0u8; bytes];
+    let mut rng = OsRng::new().map_err(|_| Error)?;
+    rng.fill_bytes(value.as_mut_slice());
+    Ok(value)
+}
+
+pub fn random_bytes<T: ArrayLength<u8>>() -> Result<GenericArray<u8, T>, Error> {
+    Ok(GenericArray::clone_from_slice(
```
lovesh: Don't you want `GenericArray` to avoid heap allocation and allocate on stack? Using `random_vec` will lead to heap allocation due to `Vec`. Why not get the mutable slice of the `GenericArray` with ` as_mut_slice` and then pass the slice to `rng.fill_bytes`?

libursa/src/encryption/symm/aescbc.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        default_impl!($name);
+        drop_impl!($name);
+        serialize_impl!($name, $visitor);
+    };
+}
+
+aes_cbc_hmac_impl!(
+    Aes128CbcHmac256,
+    U16,
```
lovesh: I see these `U` values defined in http://fizyk20.github.io/generic-array/typenum/consts/index.html but what is the objective here?

libursa/src/encryption/symm/aescbc.rs
```diff
@@ -0,0 +1,141 @@
...
+
+macro_rules! aes_cbc_hmac_impl {
+    ($name:ident, $algokeysize:ident, $keysize:ident, $noncesize:ident, $tagsize:ident, $algo:ident, $hash:ident, $visitor:ident) => {
+        #[derive(Debug, Clone, Eq, PartialEq)]
+        pub struct $name {
+            key: GenericArray<u8, $keysize>,
+        }
+
+        impl Encryptor for $name {
+            type MinSize = U48;
```
lovesh: Is `MinMinSize` always `U48`?

libursa/src/encryption/symm/aescbc.rs
```diff
@@ -0,0 +1,141 @@
...
+
+            fn new(key: GenericArray<u8, $keysize>) -> Self {
+                Self { key }
+            }
+        }
+
+        impl Aead for $name {
+            type NonceSize = $noncesize;
+            type TagSize = $tagsize;
+            type CiphertextOverhead = U0;
```
lovesh: Is `CiphertextOverhead` always `U0`?

libursa/src/encryption/symm/aescbc.rs
```diff
@@ -0,0 +1,141 @@
...
+use subtle::ConstantTimeEq;
+use zeroize::Zeroize;
+
+type Aes128Cbc = Cbc<Aes128, Pkcs7>;
+type Aes256Cbc = Cbc<Aes256, Pkcs7>;
+type HmacSha256 = Hmac<Sha256>;
+type HmacSha512 = Hmac<Sha512>;
+
+macro_rules! aes_cbc_hmac_impl {
+    ($name:ident, $algokeysize:ident, $keysize:ident, $noncesize:ident, $tagsize:ident, $algo:ident, $hash:ident, $visitor:ident) => {
```
lovesh: What is `$algokeysize`?

libursa/src/encryption/symm/aescbc.rs
```diff
@@ -0,0 +1,141 @@
...
+                Self { key }
+            }
+        }
+
+        impl Aead for $name {
+            type NonceSize = $noncesize;
+            type TagSize = $tagsize;
+            type CiphertextOverhead = U0;
+
+            fn encrypt<'msg, 'aad>(
```
lovesh: Some documentation of `encrypt` and `decrypt` would be helpful?

libursa/src/encryption/mod.rs
mikelodder7: GenericArray does not have a creation method to do that. Otherwise I would if you can find one let me know.

libursa/src/encryption/symm/aescbc.rs
mikelodder7: Read the comments in https://github.com/hyperledger/ursa/pull/76/files#diff-882b2ef604a9130ce242dd63b5458972R213

mikelodder7: Each algorithm has a min size thats fixed.

libursa/src/encryption/symm/aescbc.rs
mikelodder7: Not quite sure why its there in the `Aead` trait but everywhere I see it used, its always zero.

mikelodder7: See other comments.

libursa/src/encryption/symm/aescbc.rs
mikelodder7: Like what. Is it not clear how to use encrypt?

libursa/src/encryption/symm/aescbc.rs
mikelodder7: See other comments.

libursa/src/encryption/symm/aescbc.rs
mikelodder7: https://github.com/hyperledger/ursa/pull/76/files#diff-49cadd47a44ba93f412f87ba4cc2427fR57

libursa/src/encryption/symm/aescbc.rs
lovesh: The logic of `encrypt` and `decrypt`

libursa/src/encryption/symm/aescbc.rs
lovesh: Which comments? Can you point out where it is used?

libursa/src/encryption/symm/aescbc.rs
lovesh: The reason I asked was because I did not understand the logic. Can you point to some link (outside this PR works too) where it is described?

mikelodder7: This is just a macro. The caller doesn't know the difference. If you want to understand the logic, then look at the code where its used.

libursa/src/encryption/symm/aescbc.rs
mikelodder7: I don't think its necessary. Anyone crypto minded knows what encrypt and decrypt do.

mac-arrap:
@hartm @brentzundel could you please review so we can merge this?

PR #77, opened 2019-11-06
PS signatures: Support scheme in section 4.2 and reformulate scheme in section 6.1

lovesh:


libzmix/src/signatures/ps/keys.rs
```diff
@@ -6,56 +6,43 @@ use super::{OtherGroup, SignatureGroup};
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Sigkey {
-    pub X: SignatureGroup,
+    pub x: FieldElement,
```
mikelodder7: This is a breaking change. Any code that depends on these won't work now. Please increment the version number of libzmix to 0.2.0

libzmix/src/signatures/ps/keys.rs
```diff
@@ -6,56 +6,43 @@ use super::{OtherGroup, SignatureGroup};
...
+    pub y: Vec<FieldElement>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Verkey {
```
mikelodder7: Same as above.

lovesh:
> Is it more efficient to use this versus the Signature recipient to blind their messages and do it using the existing code?

The signature recipient blinds their messages, in this PR as well as the old code.
`Signature` is only used while signing unblinded messages. When a blind signature is needed, `BlindSignature` is used. 
There are 2 motivations for this PR:
1. The efficiency of signer since it persists all elements (`y_i`) in secret key and does not need to do a multi-scalar multiplication with `Y_i^{m_i}` for all unblinded `i`.
2. Adding another scheme from the PS paper that does not support blind signatures. This is more efficient.

PR #79, opened 2019-11-19
Update bbs vs ps benchmarks

mikelodder7:


libzmix/benches/bbs_vs_ps.rs
```diff
@@ -29,16 +37,62 @@ fn sign_messages_benchmark(c: &mut Criterion) {
...
+                pok.gen_proof(&chal).unwrap()
+            })
+        });
+
+        let pok = PSPoKOfSignature::init(&sig, &pk, &params, attributes.as_slice(), None, HashSet::new()).unwrap();
+        let chal = FieldElement::from_msg_hash(&pok.to_bytes());
+        let proof = pok.gen_proof(&chal).unwrap();
 
-fn bbs_prove_benchmark(c: &mut Criterion) {}
+        c.bench_function(label.as_str(), |b| {
```
lovesh: Why do you use `label` as the id and not something like `format!("ps verify proof {} atts", atts)` the way you do for BBS+ sigs above?

libzmix/benches/bbs_vs_ps.rs
```diff
@@ -29,16 +37,62 @@ fn sign_messages_benchmark(c: &mut Criterion) {
...
+            b.iter(|| {
+                let pok = BBSPoKOfSignature::init(&sig, &pk, attributes.as_slice(), None, HashSet::new()).unwrap();
+                let challenge = FieldElement::from_msg_hash(&pok.to_bytes());
+                pok.gen_proof(&challenge).unwrap()
+            })
+        });
+        let poks = BBSPoKOfSignature::init(&sig, &pk, attributes.as_slice(), None, HashSet::new()).unwrap();
+        let challenge = FieldElement::from_msg_hash(&poks.to_bytes());
+        let proof = poks.gen_proof(&challenge).unwrap();
+        c.bench_function(format!("bbs+ verify proof {} atts", atts).as_str(), |b| {
```
lovesh: This benchmark works on the same proof data in each iteration. Are you sure this will not use cached results? A more accurate approach will to collect proofs from previous benchmark in a vector (if you want to exclude time to `push` to a vector, you can use `iter_custom` in place of `iter` and calculate time manually inside it) like this
```
let mut proofs = vec![];
 c.bench_function(format!("bbs+ generate proof {} atts", atts).as_str(), |b| {
            b.iter(|| {
                ...
                let p = pok.gen_proof(&challenge).unwrap();
                proofs.push(p)
```
Then use `iter_batched` to pass the `proofs` to the benchmark.

libzmix/benches/bbs_vs_ps.rs
```diff
@@ -29,16 +37,62 @@ fn sign_messages_benchmark(c: &mut Criterion) {
...
+
+        c.bench_function(format!("ps generate proof {} atts", atts).as_str(), |b| {
+            b.iter(|| {
+                let pok = PSPoKOfSignature::init(&sig, &pk, &params, attributes.as_slice(), None, HashSet::new()).unwrap();
+                let chal = FieldElement::from_msg_hash(&pok.to_bytes());
+                pok.gen_proof(&chal).unwrap()
+            })
+        });
+
+        let pok = PSPoKOfSignature::init(&sig, &pk, &params, attributes.as_slice(), None, HashSet::new()).unwrap();
```
lovesh: Same comment as above about collecting proofs from previous benchmark

libzmix/benches/bbs_vs_ps.rs
mikelodder7: As a maintainer you can update this directly. The point of the benchmark is to time each component. Partially why I do it this way is to eliminate bias in the timing as well as Rust has some ownership complaints if I try stuff like your suggesting. Try it on your end and see what you can get for it

libzmix/benches/bbs_vs_ps.rs
mikelodder7: its used as the label to the params. Doesn't make much of a difference

libzmix/benches/bbs_vs_ps.rs
mikelodder7: It's not caching. If you add a println! to the signature, its different everytime

PR #80, opened 2019-11-20
Make CredentialValues and SubProofRequest structs Deserialize-able

denzenin:
#### Why am I doing this? 
I was experimenting with usage of `ursa` in iOS and during those experiments I have created my tiny `rust` wrapper for `ursa` with a different `ffi` interface that suited my goals.
While making a wrapper I found useful the option to deserialise `SubProofRequest` and `CredentialValues` structures from `JSON` as iOS application can receive already created mentioned structures in a form of `JSON`. Thus creating them through builders via `ffi` could be unnecessary.

#### What changed
Trait `Deserialize` is added to the structures:
* `SubProofRequest`
* `CredentialValues`

PS: I am not sure that it is an `ursa` way to do things, but hope such a changes won't hurt :)


mikelodder7:
Thanks for this PR. Can you add your DCO to the commit?
The easiest way to do this is to use `git commit -s --amend`

mikelodder7:
Just fix the out-of-date issue and we can merge it.

libursa/src/cl/mod.rs
```diff
@@ -150,7 +150,7 @@ impl CredentialValue {
 }
 
 /// Values of attributes from `Claim Schema` (must be integers).
-#[derive(Debug)]
+#[derive(Debug, Deserialize)]
```
lovesh: Since you are not deriving `Serialize`, how would `CredentialValues` be serialized in the first place?

libursa/src/cl/mod.rs
```diff
@@ -779,7 +779,7 @@ pub struct BlindedCredentialSecretsCorrectnessProof {
 
 /// “Sub Proof Request” - input to create a Proof for a credential;
 /// Contains attributes to be revealed and predicates.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Deserialize)]
```
lovesh: Since you are not deriving `Serialize`, how would `SubProofRequest` be serialized in the first place?

PR #84, opened 2019-11-28
Add method to get response from Schnorr protocol

lovesh:

### DIFF
--- a/libzmix/src/signatures/bbs/pok_sig.rs
+++ b/libzmix/src/signatures/bbs/pok_sig.rs
@@ -199,6 +199,23 @@ impl PoKOfSignatureProof {
         bytes
     }
 
+    /// Get the response from post-challenge phase of the Sigma protocol for the given message index `msg_idx`.
+    /// Used when comparing message equality
+    pub fn get_resp_for_message(&self, msg_idx: usize) -> Result<FieldElement, BBSError> {
+        // 2 elements in self.proof_vc_2.responses are reserved for `&signature.e` and `r2`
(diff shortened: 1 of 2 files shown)


libzmix/src/signatures/bbs/pok_sig.rs
```diff
@@ -360,4 +377,80 @@ mod tests {
...
+        let same_blinding = FieldElement::random();
+
+        let mut blindings_1 = FieldElementVector::random(message_count - 1);
+        blindings_1.insert(1, same_blinding.clone());
+
+        let mut blindings_2 = FieldElementVector::random(message_count - 1);
+        blindings_2.insert(4, same_blinding.clone());
+
+        // Blinding for the same message is kept same
+        assert_eq!(blindings_1[1], blindings_2[4]);
```
brentzundel: Do the blindings need to be the same in order for the Schnorr proof to work?  I thought it was sufficient for the holder to know the blindings for each hidden value used in the equality proof.

lovesh: The blindings need to be same when proving equality. The response for each `message` is of the form `blinding + challenge*message` where `challenge` is common across all messages (and statements). The verifier expects the response for same messages to be same.

brentzundel: How does that work with link secrets, which are blinded differently in each credential? I feel like I'm missing something obvious.

lovesh:
Link secrets are given the same blinding during proofs.

mac-arrap:
@lovesh can you please resolve merge conflicts and then please feel free to merge

PR #88, opened 2019-12-17
Cfg changes so each feature works independently

mikelodder7:
Consumers of Ursa typically want only select features and up until now, these features were all tightly coupled.

This PR fixes the dependencies without changing the working code. This makes adjustments to the compile-time settings so each feature will work without any dependencies on others. For example, a consumer may only want `ecdsa_secp256k1` and now can do so. Before, the compiler would barf if this feature was the only one enabled.


dhuseby:
@mikelodder7 I just cloned this repo and tried a few things just to sanity check. I ran this:

```
$ cd libursa && cargo build --features="ecdsa_secp256k1"
```

I had to do that in libursa because cargo doesn't allow the specification of features on the command line in the root of a workspace? Seems weird but...OK? Anyway, I immediately got some warnings and the build failed. Here's the warnings:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/user/Projects/ursa/libursa/Cargo.toml
workspace: /home/user/Projects/ursa/Cargo.toml
```

And the build failed with:

```
   Compiling ursa v0.3.1 (/home/user/Projects/ursa/libursa)
error[E0428]: the name `ecdsa_secp256k1` is defined multiple times
   --> libursa/src/signatures/secp256k1.rs:220:1
    |
113 | mod ecdsa_secp256k1 {
    | ------------------- previous definition of the module `ecdsa_secp256k1` here
...
220 | mod ecdsa_secp256k1 {
    | ^^^^^^^^^^^^^^^^^^^ `ecdsa_secp256k1` redefined here
    |
    = note: `ecdsa_secp256k1` must be defined only once in the type namespace of this module

error: aborting due to previous error

For more information about this error, try `rustc --explain E0428`.
error: could not compile `ursa`.
```

It looks like this isn't a clean patch just yet. I'll let you take another crack at it. Just for the record:

```
$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/user/.rustup

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.42.0-nightly (3a3f4a7cb 2019-12-28)
```

mikelodder7:
@dhuseby try using `cargo build --no-default-features --features=ecdsa_secp256k1`. The problem is you can't tell it to compile both ecdsa_secp256k1 and ecdsa_secp256k1_asm. That should work. However, I'll add compiler flags so you can't do both.

mikelodder7:
Pushed update to throw a compiler user friendly message now. Make sure you turn off default features

libursa/src/bn/openssl.rs
```diff
@@ -6,15 +6,16 @@ use openssl::bn::{BigNum, BigNumContext, BigNumRef, MsbOption};
...
 use openssl::hash::{hash, Hasher, MessageDigest};
 
-#[cfg(feature = "serialization")]
-use serde::ser::{Error as SError, Serialize, Serializer};
-
-#[cfg(feature = "serialization")]
-use serde::de::{Deserialize, Deserializer, Error as DError, Visitor};
+#[cfg(feature = "serde")]
+use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
+//use serde::ser::{Error as SError, Serialize, Serializer};
```
vaporos: Is the commented-out code here intentional?

mikelodder7: I just need to delete the commented code.

libursa/src/bn/rust.rs
```diff
@@ -1,23 +1,20 @@
...
-#[cfg(feature = "serialization")]
-use serde::de::{Deserialize, Deserializer, Error as DError, Visitor};
+#[cfg(feature = "serde")]
+use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
 
 use std::cmp::Ord;
 use std::cmp::Ordering;
-#[cfg(feature = "ffi")]
-use std::error::Error;
+//#[cfg(feature = "ffi")]
```
brentzundel: why comment out rather than remove?

libursa/src/wasm/mod.rs
```diff
@@ -1,19 +1,20 @@
...
 pub mod ed25519;
+#[cfg(feature = "encryption")]
 pub mod encryption;
+#[cfg(feature = "ecdsa_secp256k1")]
 pub mod secp256k1;
 
-//#[cfg(feature = "cl")]
-//pub mod cl;
-
+//#[cfg(any(feature = "ed25519", feature = "ecdsa_secp256k1"))]
```
brentzundel: why commented out and not deleted?

PR #89, opened 2019-12-27
remove useless cargo-lock

HaoXuan40404:
This file is useless, and maybe cause some problems when build release, so perhaps delete it would be better?
Signed-off-by: HaoXuan40404 <444649358@qq.com>


danintel:
According to this:
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
Cargo.lock files should be checked in for applications, but not libraries.
OTOH, it does no harm as it is ignored in the case of libraries.

HaoXuan40404:
> According to this:
> https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
> Cargo.lock files should be checked in for applications, but not libraries.
> OTOH, it does no harm as it is ignored in the case of libraries.

Yes, but when I build this repo, Cargo.lock will change. Maybe this will make confused?

dcmiddle:
Please see discussion here: https://github.com/hyperledger/ursa/pull/61

PR #91, opened 2020-01-17
Upgrade aead/aes-gcm/chacha20poly1305 crates

tarcieri:
Updates the following crates:

- `aead` => v0.2
- `aes-gcm` => v0.3.0
- `chacha20poly1305` => v0.3.1

The `aead` crate v0.2 adds mandatory in-place/detached APIs. Perhaps these should be split into an e.g. `AeadInPlace` trait, however for now they are stubbed out with `unimplemented!()`.


tarcieri:
@mikelodder7 done

PR #92, opened 2020-01-24
IS-1222: Better error in case proof request uses attr name in both attrs and preds

Artemkaaas:
Signed-off-by: artem.ivanov <artem.ivanov@dsr-company.com>


Artemkaaas:
Related issue:https://jira.hyperledger.org/browse/IS-1222

PR #96, opened 2020-02-10
Fix build error (missing 'secret' macro) when using 'cl' feature without encryption

andrewwhitehead:


andrewwhitehead:
@lovesh Running `cargo check` seems to be a faster and less resource intensive way to identify this issue: `cargo check --no-default-features --features=cl` from the libursa directory.

lovesh:
@andrewwhitehead Yes, that works as well.
Are you suggesting that we run the tests with all features so that the build is created only once but run `cargo check` for each feature (like `cargo check --no-default-features --features=<feature name>`) to ensure that each feature can be built?

andrewwhitehead:
@lovesh Not exactly, the ed25519 and secp256k1 cases have test suites so you'd need to build the binaries anyway I think (with the appropriate flags). By running them in the same pod you're only downloading rust once and only downloading and building the dependencies once, though.

lovesh:
@andrewwhitehead I was referring to the library tests, sorry for being unclear. I don't know what a "pod" is in Azure CI lingo but I think you mean a stage and yes Rust should be downloaded only twice (stable and nightly), probably in the first stage, followed by `cargo check`s for features in the next stage, followed by stages for build and tests.
I am puzzled by looking at `azure-pipelines.yml`, as its downloading Rust for each job? Shouldn't it be downloaded in a separate stage? @mikelodder7

libursa/src/utils/macros.rs
```diff
@@ -17,6 +17,20 @@ macro_rules! array_copy {
     };
 }
 
+#[cfg(any(
```
lovesh: Wouldn't this feature set work for only `impl_bytearray`? You need this feature set for `secret` macro.

libursa/src/utils/macros.rs
```diff
@@ -17,6 +17,20 @@ macro_rules! array_copy {
...
+    feature = "ed25519_asm",
+    feature = "ecdh_secp256k1",
+    feature = "ecdh_secp256k1_native",
+    feature = "ecdh_secp256k1_asm",
+    feature = "ecdsa_secp256k1",
+    feature = "ecdsa_secp256k1_native",
+    feature = "ecdsa_secp256k1_asm",
+    feature = "x25519",
+    feature = "x25519_asm",
+    feature = "wasm"
```
lovesh: You need to add the feature `cl` in this list.

andrewwhitehead: I don't think impl_bytearray is used in the cl feature. It compiles without it.

libursa/src/utils/macros.rs
andrewwhitehead: Yes, that's the idea. `secret` isn't gated by additional features, only the features on the whole module.

lovesh: I misunderstood. I thought you were trying to make `secret!` gated.

PR #103, opened 2020-03-16
Check public key length

mikelodder7:
Check for public key length before converting

### DIFF
--- a/libursa/src/signatures/ed25519.rs
+++ b/libursa/src/signatures/ed25519.rs
@@ -22,6 +22,8 @@ impl Ed25519Sha512 {
     pub fn ver_key_to_key_exchange(pk: &PublicKey) -> Result<PublicKey, CryptoError> {
         use curve25519_dalek::edwards::CompressedEdwardsY;
 
+        // Verify its a valid public key
+        PK::from_bytes(&pk[..]).map_err(|e| CryptoError::ParseError(e.to_string()))?;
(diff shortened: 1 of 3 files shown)


libursa/src/signatures/ed25519.rs
```diff
@@ -22,6 +22,9 @@ impl Ed25519Sha512 {
     pub fn ver_key_to_key_exchange(pk: &PublicKey) -> Result<PublicKey, CryptoError> {
         use curve25519_dalek::edwards::CompressedEdwardsY;
 
+        if pk.len() != PUBLIC_KEY_SIZE {
```
Artemkaaas: In my opinion it rather `PublicKey` has to contain the method `new or from_bytes`() and do length checking there.
Else you have to put the same check in every function using PublicKey

mikelodder7: `PublicKey` is a generic class that can be variable in length depending on which Crypto Primitive is using it. So it must be done here.

libursa/src/signatures/ed25519.rs
```diff
@@ -22,6 +22,8 @@ impl Ed25519Sha512 {
     pub fn ver_key_to_key_exchange(pk: &PublicKey) -> Result<PublicKey, CryptoError> {
         use curve25519_dalek::edwards::CompressedEdwardsY;
 
+        // Verify its a valid public key
```
brentzundel:
```suggestion
        // Verify it's a valid public key
```

PR #104, opened 2020-03-18
Upgrades version of ed25519-dalek

eadventurous:
## Description of changes
Upgrade from 1.0.0-pre.2 to 1.0.0-pre.3
And corresponding fixes for other lib versions and src files

## Benefits
There is an issue with the usage of latest ursa (0.3.2) with latest libp2p(0.16.2): they have incompatible versions of ed25519-dalek (ursa - 1.0.0-pre.2 and libp2p - 1.0.0-pre.3). It could be solved only by downgrading libp2p to version 0.13.1 which is rather old now.

Therefore this PR is meant to fix that and make latest ursa and libp2p compatible, as they have rather intersecting user groups in my opinion.

Signed-off-by: Egor Ivkov <e.o.ivkov@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,7 +94,7 @@ dependencies = [
  "byteorder",
  "lazy_static",
  "miracl_amcl",
- "rand 0.7.3",
+ "rand 0.7.0",
  "rayon",
  "serde",
  "serde_derive",
@@ -166,7 +166,7 @@ dependencies = [
  "hex",
  "hkdf",
  "pairing-plus",
- "rand 0.7.3",
+ "rand 0.7.0",
  "rayon",
  "serde",
  "serde-wasm-bindgen",
@@ -278,7 +278,7 @@ dependencies = [
  "failure",
  "lazy_static",
  "merlin",
- "rand 0.7.3",
+ "rand 0.7.0",
  "serde",
  "serde_derive",
  "serde_json",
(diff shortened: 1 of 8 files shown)


eadventurous:
Seems like this upgrade is blocked by rust-num/num-bigint#128 for now. It is needed to upgrade the version of num-bigint from `0.2` to `0.3`.

mikelodder7:
> Seems like this upgrade is blocked by [rust-num/num-bigint#128](https://github.com/rust-num/num-bigint/issues/128) for now. It is needed to upgrade the version of num-bigint from `0.2` to `0.3`.

`num-bigint` 0.3 hasn't been released. The highest it goes is 0.2.6 and that version is still tied to 0.5. I'm tempted to push my own crate that does work so this can be unblocked. I also need to update glass_pumpkin but have been stuck on this version as well.  Perhaps I'll do that later today.

eadventurous:
The `0.3` version of `num-bigint` was released on 12th of June. So hopefully we can also upgrade the versions here

eadventurous:
The `asm` build target seems to also fail on previous master branch commits with the same error: `error: the legacy LLVM-style asm! syntax is no longer supported`. So I guess this PR does not introduce new problems for `asm`.

mikelodder7:
There’s already a PR that needs to be merged that fixes the asm build error

eadventurous:
> There’s already a PR that needs to be merged that fixes the asm build error

So I think it is possible to merge this PR after the fix for the `asm` will be merged?

mikelodder7:
Most likely. For now, I don't foresee any problem.

PR #105, opened 2020-03-19
Iterate tails indices in order for better cache locality

andrewwhitehead:


mikelodder7:
Is there a noticeable performance difference switching from a HashMap to BTreeMap?

JAG-UK:
> Is there a noticeable performance difference switching from a HashMap

Anecdotally yes:
https://users.rust-lang.org/t/hashmap-vs-btreemap/13804/2

  | BTreeMap | HashMap
-- | -- | --
parse canada.json | 10.9ms | 11.9ms
stringify canada.json | 11.3ms | 11.6ms
parse citm_catalog.json | 5.5ms | 11.2ms
stringify citm_catalog.json | 1.4ms | 3.3ms
parse twitter.json | 2.5ms | 2.9ms
stringify twitter.json | 0.6ms | 0.9ms

andrewwhitehead:
The BTreeMap is likely microseconds slower to create, but it may have an impact when accessing larger tails files with slow IO or a smaller cache (no difference on my machine). When creating the Witness both the issuer and prover are currently walking through the tails file in essentially random order.

PR #106, opened 2020-03-25
Implement try_clone for CredentialKeyCorrectnessProof

andrewwhitehead:


brentzundel:
I don't have a problem merging this PR, but am uncomfortable doing so while there are build errors.
@mikelodder7 do you know what is happening there?

mikelodder7:
> I don't have a problem merging this PR, but am uncomfortable doing so while there are build errors.
> @mikelodder7 do you know what is happening there?

I believe its because we need to update Cargo.lock. If they update that, it should fix it.

PR #107, opened 2020-04-04
Fix warnings and update to 0.3.5

mikelodder7:
This updates bulletproofs to use amcl_wrapper 0.3.5 and removes lots of warnings

### DIFF
--- a/libzmix/bulletproofs_amcl/src/r1cs/verifier.rs
+++ b/libzmix/bulletproofs_amcl/src/r1cs/verifier.rs
@@ -198,6 +198,7 @@ impl<'a> Verifier<'a> {
         (wL, wR, wO, wV, wc)
     }
 
+    // Only used for debugging purposes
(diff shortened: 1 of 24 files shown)


libzmix/bulletproofs_amcl/src/r1cs/verifier.rs
```diff
@@ -198,53 +198,53 @@ impl<'a> Verifier<'a> {
...
-                    Variable::One() => {
-                        wc -= coeff;
-                    }
-                }
-            }
-        }
-
-        (WL, WR, WO, WV, wc)
-    }
+    // #[cfg(test)]
```
brentzundel, line 201:
```suggestion
    // Only used for debugging purposes
    // #[cfg(test)]
```

brentzundel:
@mac-arrap or @danintel can we get another set of eyes on this one?

PR #111, opened 2020-04-08
Add non-extended chacha20poly1305

andrewwhitehead:
Also updates chacha20poly1305 crate to 0.4.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,13 +353,14 @@ dependencies = [
 
 [[package]]
 name = "chacha20poly1305"
-version = "0.3.3"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26ffa408231820dc95b7162a0b7548ab6ae0373be660a8e9ccad2a8b570a3870"
+checksum = "48901293601228db2131606f741db33561f7576b5d19c99cd66222380a7dc863"
 dependencies = [
  "aead",
  "chacha20",
  "poly1305",
+ "stream-cipher",
  "zeroize",
 ]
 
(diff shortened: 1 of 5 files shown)


mikelodder7:
@andrewwhitehead Thanks for this PR. Is there a compatibility reason or streaming reason for using chacha20poly1305? In most cases xchacha20poly1305 is preferred except in those two cases.

andrewwhitehead:
@mikelodder7 Yes, it's for compatibility with the current Indy pack/unpack, which advertises xchacha but doesn't actually use it.

kdenhartog:
> 
> 
> @mikelodder7 Yes, it's for compatibility with the current Indy pack/unpack, which advertises xchacha but doesn't actually use it.

Yeahhhhh.... Sorry bout that headache. Maybe we can make it consistent eventually?

andrewwhitehead:
@kdenhartog  Because the nonce is detached, it would be pretty easy to support both by checking its length first (12 or 24 bytes).

andrewwhitehead:
Also the updated pack/unpack exists here now:
https://github.com/bcgov/indy-shared-rs/blob/master/indy-utils/src/pack/alg.rs

It uses the crypto_box crate to create/open the nacl crypto box, and implements the sealed box manually.

kdenhartog:
> @kdenhartog Because the nonce is detached, it would be pretty easy to support both by checking its length first (12 or 24 bytes).

We could just add support for both rather than relying on the implementation supporting only xchacha as well. I think it would be good to not indicate the wrong type of algorithm being used in the headers if it's possible to get everyone doing that.

PR #112, opened 2020-04-08
Add method to derive x25519 from an ed25519 private key

andrewwhitehead:
This method is compatible with libsodium's `crypto_sign_ed25519_sk_to_curve25519`.

This PR also adds `Ed25519Sha519::keypair_from_secret` which is compatible with libsodium's `crypto_sign_seed_keypair`. @mikelodder7  I didn't add this as a separate KeyGenOption as it didn't seem to be generally applicable, but I'll let you be the judge.

### DIFF
--- a/libursa/src/signatures/ed25519.rs
+++ b/libursa/src/signatures/ed25519.rs
@@ -56,8 +56,8 @@ impl Ed25519Sha512 {
             return Err(CryptoError::ParseError(format!("Invalid secret provided")));
         }
         let mut private = vec![0u8; 64];
-        private[0..32].copy_from_slice(seed);
-        let sk = SK::from_bytes(&private[..32]).unwrap();
+        private[..32].copy_from_slice(&seed[..32]);
+        let sk = SK::from_bytes(&seed[..32]).unwrap();
         let pk = PK::from(&sk).to_bytes().to_vec();
         private[32..].copy_from_slice(pk.as_ref());
         Ok((PublicKey(pk), PrivateKey(private)))
--- a/libursa/src/signatures/ed25519.rs
+++ b/libursa/src/signatures/ed25519.rs
@@ -1,7 +1,7 @@
 pub const ALGORITHM_NAME: &str = "ED25519_SHA2_512";
 
 use super::{KeyGenOption, SignatureScheme};
-use ed25519_dalek::{Keypair, PublicKey as PK, Signature};
+use ed25519_dalek::{Keypair, PublicKey as PK, SecretKey as SK, Signature};
 pub use ed25519_dalek::{
     EXPANDED_SECRET_KEY_LENGTH as PRIVATE_KEY_SIZE, PUBLIC_KEY_LENGTH as PUBLIC_KEY_SIZE,
     SIGNATURE_LENGTH as SIGNATURE_SIZE,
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
Yes, as much as I can. The last test is deriving a common Indy keypair.

andrewwhitehead:
@mikelodder7 Tested here: https://github.com/andrewwhitehead/indy-sdk/tree/test/derive-key

mikelodder7:
@andrewwhitehead I'm not seeing that function in that branch in that file. Can you provide the direct link?

andrewwhitehead:
@mikelodder7 It's over in https://github.com/andrewwhitehead/indy-sdk/blob/test/derive-key/libindy/tests/crypto.rs

The whole test suite runs anyway, apart from some unrelated pool genesis transaction tests. `indy_crypto_sign_works` might be more relevant as it's generating a keypair from a seed, signing data and comparing the result to an expected value.

kdenhartog:
The function that does the bilateral mapping by X25519 can be found here: https://github.com/dalek-cryptography/curve25519-dalek/blob/17698df9d4c834204f83a3574143abacb4fc81a5/src/edwards.rs#L473

On Indy-sdk side I'm having a hard time figuring out which sodiumoxide function is being called to then see what libsodium fn is called, so I can validate the same math. I'm getting lost here: https://github.com/hyperledger/indy-sdk/blob/be0618bd40931e50ee64cbdb0f640f9a8f36852a/libindy/indy-utils/src/crypto/ed25519_sign/sodium.rs#L29

Is anyone else able to figure out how this is working?

mikelodder7:
@andrewwhitehead I see the test but what I'm interested in is whether this code you are pushing to Ursa is functionally the same. I actually think this code could be moved to Indy and kept out of Ursa since this is an Indy specific implementation.

kdenhartog:
> 
> 
> @andrewwhitehead I see the test but what I'm interested in is whether this code you are pushing to Ursa is functionally the same. I actually think this code could be moved to Indy and kept out of Ursa since this is an Indy specific implementation.

This is quite common functionality in libraries. From what I know, libsodium, monocipher, and dalek-cryptography currently support this. I see it as beneficial to be kept within ursa.

andrewwhitehead:
I think the `sign_key_to_key_exchange` method is pretty much essential in order to do anything useful with `verkey_to_key_exchange` which is already in there.

I don't think `keypair_from_secret` can be implemented in Indy without adding `ed25519_dalek` as an explicit dependency on top of Ursa (or rewriting it). It's also not an Indy-specific method for creating the secret key, it just expands the secret part to a full keypair.

mikelodder7:
@andrewwhitehead and I met over zoom to discuss the details. I'm happy with it now.

libursa/src/signatures/ed25519.rs
```diff
@@ -34,6 +34,36 @@ impl Ed25519Sha512 {
             ))),
         }
     }
+
```
brentzundel, line 37: Please add comments here to document the purpose of the sign_key_to_key_exchange function and how to use it.

libursa/src/signatures/ed25519.rs
```diff
@@ -34,6 +34,36 @@ impl Ed25519Sha512 {
...
+        }
+        // hash secret
+        let hash = sha2::Sha512::digest(&sk[..32]);
+        let mut output = [0u8; 32];
+        output.copy_from_slice(&hash[..32]);
+        // clamp result
+        let secret = x25519_dalek::StaticSecret::from(output);
+        Ok(PrivateKey(secret.to_bytes().to_vec()))
+    }
+
```
brentzundel, line 53: Please add comments here to document the purpose of the expand_keypair function and how to use it

PR #114, opened 2020-04-14
Deterministic bbs

mikelodder7:
This PR does the following:

1- Separates BBS+ into its own sub-crate
2- Adds a deterministic public key using hash2curve
3- Adds helper functions for doing issuing and proving


libzmix/bbs/src/keys.rs
```diff
@@ -0,0 +1,238 @@
...
+/// a commitment to the secret key `w`
+/// and a generator for each message in `h`
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct PublicKey {
+    /// Blinding factor generator
+    pub h0: G1, //blinding factor base
+    /// Base for each message to be signed
+    pub h: Vec<G1>, //base for each message to be signed
+    /// Commitment to the private key
+    pub w: G2, //commitment to private key
```
brentzundel, line 42:
```suggestion
    /// Blinding factor generator
    pub h0: G1,
    /// Base for each message to be signed
    pub h: Vec<G1>,
    /// Commitment to the private key
    pub w: G2,
```

libzmix/bbs/src/signature.rs
```diff
@@ -0,0 +1,363 @@
...
+    pub a: G1,
+    /// e
+    pub e: SignatureNonce,
+    /// s
+    pub s: SignatureNonce,
+}
+
+impl BlindSignature {
+    /// 1 or more messages have been hidden by the signature recipient. The remaining
+    /// known messages are in `messages`. To which generator they correspond is in `message_indices`.
```
brentzundel, line 77:
```suggestion
    /// known messages are in `messages`. The generator to which they correspond is in `message_indices`.
```

PR #116, opened 2020-04-22
Implement compressed form

mikelodder7:
This PR updates to a newer version of amcl_wrapper as well as implements a compressed form for keys and proofs. 

For example, points can be stored as just the x-coordinates. Curve order elements can remove leading zero bytes.

The result is that points in G1 drop to 48 bytes, G2 drop to 96 bytes, and elements in Zq drop to 32 bytes.

The major gains come from proof sizes. A proof with 5 hidden messages is 933 bytes in uncompressed form and 530 in compressed form.

Public Key sizes go down to 1/2.

Signatures go from 193 bytes to 112 bytes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,6 +76,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ee5cca1ddc8b9dceb55b7f1272a9d1e643d73006f350a20ab4926d24e33f0f0d"
 
 [[package]]
+name = "amcl-milagro"
+version = "3.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9a7f1be21d1876ce01fe62e2984fd4617fccc251f17d25098e869fd83f71e770"
+
+[[package]]
 name = "amcl_wrapper"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


brentzundel:
> > I am also not sure about introducing wire-level concerns to ursa. It feels like the wrong layer for such things to me.
> 
> Except this isn't sending anything over the wire. It's just implementing a compressed form of the points and scalars. This can be used for wire or storage or anything else. Callers can use either form as they see fit. This is the right place for it. Callers shouldn't have to know the internal representation.

It is not sending things over the wire, but it is formatting objects for external transmission or storage, and I'm not convinced that belongs inside of ursa. 

Is the compression mechanism recognized by any other software libraries, or does using the compressed forms of the points and scalars require the users on both ends to use Ursa?

mikelodder7:
> It is not sending things over the wire, but it is formatting objects for external transmission or storage, and I'm not convinced that belongs inside of ursa.

No its not doing any format. What its doing is instead of storing both the x and y values for G1
its just storing x and sign of y. Y can be computed from just x.

For field elements where there are 384 bits and only 256 bits are needed, I just chop off the leading zeros and restore them back when deserializing. There's no format specific stuff going on here.  See https://github.com/zkcrypto/pairing/tree/master/src/bls12_381 for more information about this.

brentzundel:
> No its not doing any format. What its doing is instead of storing both the x and y values for G1
> its just storing x and sign of y. Y can be computed from just x.

This is probably fine for key values, if it were documented properly what was happening.
However, I'm seeing compressed forms of:
1. public keys
1. blind signature contexts
1. proof requests
1. signature proofs
1. pok signature proofs

Now perhaps what each of those objects is doing is simply using the short form key values internally. If that is all that is happening, that is not clear. If more than that is happening, I remain concerned.

mikelodder7:
> Now perhaps what each of those objects is doing is simply using the short form key values internally
Yep, that's all its doing. If its a point, it uses just outputs the X. If it's a curve order element, it's just chopping the leading zeros. That's it.

PR #117, opened 2020-05-06
BBS+ demo

brentzundel:
Adds a demo of the full flow of credentials using bbs+ signatures.
A Prover obtains credentials from two Issuers that include the Prover's link secret.
A Verifier requests that the Prover reveal one claim from one credential, and prove the equality of one claim from each credential.


mac-arrap:
@brentzundel could you fix the merge conflicts here? I will merge after this is done

brentzundel:
@mac-arrap It should be ready to go.

PR #119, opened 2020-05-11
Update to use pairing_plus vs milagro

mikelodder7:
Sorry, this looks like a BIG PR which it is, but the biggest change is switching from Apache Milagro which is no longer under development to [pairing-plus](https://crates.io/crates/pairing-plus).
Because of this change, it touched 90% of the BBS+ code. This PR only touches the BBS+ code. I have not applied it to the rest of ZMix yet.

Using pairing-plus offers the following advantages:

1- It's under active development
2- It's slightly faster overall about 5-10% faster than milagro in every case.
3- It's significantly faster for HashToCurve, about 100X faster.
4- Built-in compression methods rather than manually doing it.
5- ABI agnostic, it runs the same speed on 64-bit and 32-bit.
6- Unified function names i.e. `to_bytes_compressed_form` everywhere vs just in BBS

### DIFF
--- a/libzmix/bbs/src/pok_vc.rs
+++ b/libzmix/bbs/src/pok_vc.rs
@@ -10,18 +10,20 @@
 //! During response generation `ProverCommitted` is consumed to create `Proof` object containing the commitments and responses.
 //! `Proof` can then be verified by the verifier.
 
-use crate::prelude::*;
+use crate::{hash_to_fr, multi_scalar_mul_const_time_g1, ProofChallenge, SignatureMessage, GeneratorG1, ToVariableLengthBytes, FR_COMPRESSED_SIZE, G1_COMPRESSED_SIZE, G1_UNCOMPRESSED_SIZE, Commitment};
 
-use amcl_wrapper::constants::{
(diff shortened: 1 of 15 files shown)


libzmix/bbs/src/pok_vc.rs
```diff
@@ -537,44 +472,33 @@ macro_rules! test_PoK_VC {
...
-    2 * FIELD_ORDER_ELEMENT_SIZE
-);
+// pok_vc_impl!(
+//     ProverCommittingG1,
+//     ProverCommittedG1,
+//     ProofG1,
+//     G1,
+//     G1_UNCOMPRESSED_SIZE,
+//     G1_COMPRESSED_SIZE
+// );
```
brentzundel, line 496: Why is this commented out and not simply removed?

libzmix/bbs/src/pok_vc.rs
```diff
@@ -584,21 +508,21 @@ mod tests {
...
+    //     let n = 5;
+    //     test_PoK_VC!(
+    //         n,
+    //         ProverCommittingG2,
+    //         ProverCommittedG2,
+    //         ProofG2,
+    //         G2,
+    //         G2_UNCOMPRESSED_SIZE
+    //     );
+    // }
```
brentzundel, line 527: Why is this commented out and not simply removed?

PR #120, opened 2020-05-20
Verifier hash

brentzundel:
add convenience function to verifier - create_challenge_hash
modify tests to use convenience function
fix inputs of prover::create_challenge_hash to match (slices not vectors)
add tests

### DIFF
--- a/libzmix/bbs/src/prover.rs
+++ b/libzmix/bbs/src/prover.rs
@@ -123,8 +123,8 @@ impl Prover {
     /// * `nonce` - a SignatureNonce
     /// * `claims` - a vec of strings the prover wishes to include in the challenge (may be empty)
     pub fn create_challenge_hash(
-        pok_sigs: Vec<PoKOfSignature>,
-        claims: Vec<&str>,
+        pok_sigs: &[PoKOfSignature],
+        claims: &[&[u8]],
         nonce: &ProofNonce,
     ) -> Result<ProofChallenge, BBSError> {
         let mut bytes = Vec::new();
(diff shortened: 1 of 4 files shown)


libzmix/bbs/src/verifier.rs
```diff
@@ -71,4 +71,35 @@ impl Verifier {
...
+    ///
+    /// # Arguments
+    /// * `proofs` - a vec of SignatureProof objects
+    /// * `proof_requests` - a corresponding vec of ProofRequest objects
+    /// * `nonce` - a SignatureNonce
+    /// * `claims` - a vec of strings the prover wishes to include in the challenge (may be empty)
+    pub fn create_challenge_hash(
+        proofs: &[SignatureProof],
+        proof_requests: &[ProofRequest],
+        claims: &[&[u8]],
```
mikelodder7, line 85: Maybe this should be optional. I'd like to use the method but I don't necessarily have any additional claims to add.

`claims: Option<&[&[u8]]>`

libzmix/bbs/src/prover.rs
```diff
@@ -123,8 +123,8 @@ impl Prover {
     /// * `nonce` - a SignatureNonce
     /// * `claims` - a vec of strings the prover wishes to include in the challenge (may be empty)
```
kdenhartog, line 124: This comment may be updated now? I think it's supposed to be an optional slice of bytes. Not recognizing if these types are the same. I'm a bit "rusty" at reading rust

image

brentzundel: good catch, fixed in [df1086c](https://github.com/hyperledger/ursa/pull/120/commits/df1086cfaad044d82c79eb6a1f02749c3a50f2e7)

brentzundel:
@mikelodder7 any other changes you'd like to see?

mikelodder7:
> @mikelodder7 any other changes you'd like to see?

Nope its all good now.

PR #127, opened 2020-06-23
Fixed ASM build

mikelodder7:
This PR fixes building with the `asm` flag. I previously had to disable it due to the compiler issue. This fixes that and reenables the pipeline run too.

### DIFF
--- a/docker/ubuntu/ursa-fossa.dockerfile
+++ b/docker/ubuntu/ursa-fossa.dockerfile
@@ -42,5 +42,5 @@ RUN cd /usr/lib/x86_64-linux-gnu \
     && make install \
     && cd .. \
     && rm -rf libsodium-1.0.18 \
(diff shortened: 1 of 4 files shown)


docker/ubuntu/ursa-fossa.dockerfile
```diff
@@ -41,5 +42,5 @@ RUN cd /usr/lib/x86_64-linux-gnu \
     && make install \
     && cd .. \
     && rm -rf libsodium-1.0.18 \
-    && curl https://sh.rustup.rs -sSf | sh -s -- -y
+    && curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2020-04-12
```
brentzundel, line 45: what requires nightly and do you know how far away those things are from being supported on stable?

mikelodder7: We test for using SIMD instructions with some of the hashes which only work with nightly.

PR #142, opened 2020-09-01
Check for zero byte and points at infinity keys

mikelodder7:
This PR checks for exceptional behavior in keys and values that could potentially lead to disastrous results like validating every signature even though I don't think this formally breaks the security of the scheme.

This could occur if the secret key is zero bytes which would result in a public key equal to the identity point. 

The problem with allowing the identity point as a public key is that every signature under this public key is also the identity point, which means that a signer falsifies the message they signed.

It also checks makes sure random values are never zero.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,9 +11,9 @@ dependencies = [
 
 [[package]]
 name = "adler"
-version = "0.2.2"
+version = "0.2.3"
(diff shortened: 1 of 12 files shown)


hartm (approved):
This looks fine (and thanks Mike!), but maybe in the future we should separate the crate updates from the core crypto code stuff?

PR #143, opened 2020-09-02
Allow to pass pre-generated safe prime numbers to Credential Definition creation

wulfraem:
To allow speeding up credential definition creation we wanted to create the safe prime numbers beforehand with `ursa::helpers::generate_safe_prime` and hold onto them until the actual credential definition is created.

In our use case we have to create multiple credential definitions at a time and having a bunch of safe primes ready for this improves performance of this a lot. :)

### DIFF
--- a/libursa/src/cl/issuer.rs
+++ b/libursa/src/cl/issuer.rs
@@ -76,31 +76,73 @@ impl Issuer {
         let (p_pub_key, p_priv_key, p_key_meta) =
             Issuer::_new_credential_primary_keys(credential_schema, non_credential_schema)?;
 
-        let (r_pub_key, r_priv_key) = if support_revocation {
-            Issuer::_new_credential_revocation_keys()
-                .map(|(r_pub_key, r_priv_key)| (Some(r_pub_key), Some(r_priv_key)))?
-        } else {
-            (None, None)
-        };
+        Issuer::_new_credential_def_with_keys(
+            credential_schema,
+            support_revocation,
+            p_pub_key,
+            p_priv_key,
+            p_key_meta,
+        )
+    }
 
-        let cred_pub_key = CredentialPublicKey {
-            p_key: p_pub_key,
-            r_key: r_pub_key,
-        };
-        let cred_priv_key = CredentialPrivateKey {
-            p_key: p_priv_key,
-            r_key: r_priv_key,
-        };
-        let cred_key_correctness_proof = Issuer::_new_credential_key_correctness_proof(
-            &cred_pub_key.p_key,
-            &cred_priv_key.p_key,
-            &p_key_meta,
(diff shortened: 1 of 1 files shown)


brentzundel (changes requested):
This PR concerns me deeply. It goes against the stated ideals of simple usability, and allows any numbers to be submitted as "primes," which could be a massive footgun.
We need to discuss this much more before merging this.

libursa/src/cl/issuer.rs
```diff
@@ -780,7 +822,33 @@ impl Issuer {
         CredentialPrimaryPublicKeyMetadata,
     )> {
         trace!(
-            "Issuer::_new_credential_primary_keys: >>> credential_schema: {:?}",
+            "Issuer::_new_credential_primary_keys_with_primes: >>> credential_schema: {:?}",
```
brentzundel, line 825:
```suggestion
            "Issuer::_new_credential_primary_keys: >>> credential_schema: {:?}",
```

brentzundel:
@wulfraem 
We discussed this on the Ursa call today, and would like to make it possible for this to happen. We are grateful for contributions and apologize for not responding more fully until now.

Here are the key concerns around your proposed changes:

1. Lack of input validation for the prime values
1. A reduction in clarity for the API
1. The library is currently undergoing significant refactoring.

In greater detail: 

1. At a minimum, the input values (the two safe primes) would need to be checked for:
    1. primality - Since there is no way to guarantee that input values were generated using `ursa::helpers::generate_safe_prime`, there should be primality checking of the input values. We anticipate this may duplicate some of the prime-checking that has already occurred for valid primes, but it would also add protection for unwary users.
    1. valid range - The call to `generate_safe_primes` currently expects a size. There would need to be a check to insure that the input values don't exceed this range.
1. We are concerned that the changed API surface may be confusing to users. Though your added comments are helpful, we feel there could be more done to guide users to the proper understanding of which function best suits their needs.
1. The ursa library is currently experiencing a large refactoring effort. This is more informational for you than anything else. Because of this refactoring, changes you propose may meet with greater than usual need to make the PR up-do-date with the master branch and deal with conflicts.

Please respond with any questions you may have. We also welcome you to join our biweekly call to discuss this PR with the group directly. You should be able to find meeting time and connection information on the [hyperledger calendar](https://wiki.hyperledger.org/display/HYP/Calendar+of+Public+Meetings)

Alexis-Falquier:
@brentzundel are there any updates to this PR? Is the plan to merge it?

brentzundel:
@Alexis-Falquier I have not heard a response from the OP about our concerns. Until those are addressed, I do not support merging this PR.

brentzundel:
This PR has been dormant for over a year, with no response from @wulfraem 
Closing

PR #146, opened 2020-09-22
Create rust.yml

dhuseby:
This will set up a basic GH action to build Ursa. It will likely need some revisions to get it to work properly since I suspect we'll need special commands for installing libsodium-dev and other build dependencies.


dhuseby:
nvrmnd. it seems like AZP is already set up. doh!

PR #162, opened 2020-11-09
Upgrade aes/aes-gcm/block-modes/chacha20poly1305 crates

tarcieri:
Updates the following crates:

- `aes` => v0.6
- `aes-gcm` => v0.8
- `block-modes` => v0.7
- `chacha20poly1305` => v0.7

Signed-off-by: Tony Arcieri <bascule@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,46 +32,46 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.5.0"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd2bc6d3f370b5666245ff421e231cba4353df936e26986d2918e61a8fd6aef6"
+checksum = "884391ef1066acaa41e766ba8f596341b96e93ce34f9a43e7d24bf0a0eaf0561"
 dependencies = [
  "aes-soft",
  "aesni",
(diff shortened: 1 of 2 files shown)


tarcieri:
Note: the `hkdf`, `hmac`, `sha2`, and `sha3` crates are all out-of-date as well, however I'm unable to update those until the `libsecp256k1` crate releases a new version.

You might consider the `k256` crate as an alternative pure Rust implementation of secp256k1. It uses more modern (projective) formulas and provides better performance: https://crates.io/crates/k256

mikelodder7:
Thanks so much, @tarcieri. I'll make the change to `k256` since it closely matches `p256` and would allow ECDSA with either curve.

mikelodder7 (approved):

dcmiddle:
> Note: the `hkdf`, `hmac`, `sha2`, and `sha3` crate dependencies are all out-of-date as well, however you'll be unable to update those until the `libsecp256k1` crate releases a new version.
> 
> You might consider the `k256` crate as an alternative pure Rust implementation of secp256k1. It uses more modern (projective) formulas and provides better performance: https://crates.io/crates/k256

Cool. I created #163 to discuss.

dcmiddle (approved):

PR #165, opened 2020-11-13
Add secret sharing schemes

mikelodder7:
Signed-off-by: Michael Lodder <redmike7@gmail.com>


ursa_sharing/src/pedersen.rs
```diff
@@ -0,0 +1,232 @@
...
+        if rhs.is_zero() {
+            Ok(())
+        } else {
+            Err(SharingError::ShareInvalidValue)
+        }
+    }
+
+    /// Reconstruct a secret from shares created from `split_secret`. The shares should be verified
+    /// first by calling `verify_share`. This method assumes all the shares have been verified.
+    /// Usually `verify_share` is called when the share is received.
```
brentzundel, line 149:
```suggestion
    /// Reconstruct a secret from shares created from `split_secret`.
    /// The shares should be verified first by calling `verify_share`.
    /// This method assumes all the shares have been verified.
    /// Usually `verify_share` is called when the share is received.
```

brentzundel (commented):
Some minor editorial suggestions.

Is there a reason the tests for each secret sharing scheme are combined? Seems like that might make adding more schemes difficult in the future.

ursa_sharing/src/pedersen.rs
```diff
@@ -0,0 +1,232 @@
...
+            return Err(SharingError::ShareLimitLessThanThreshold);
+        }
+        if threshold < 2 {
+            return Err(SharingError::ShareMinThreshold);
+        }
+        Ok(Self { threshold, limit })
+    }
+
+    /// Create shares from a secret
+    /// Caller can optionally supply a random generators for use
```
hartm, line 44:
```suggestion
    /// Caller can optionally supply random generators for use
```
Grammar police!

ursa_sharing/src/pedersen.rs
hartm, line 149: Is there a reason not to automatically call verify_share when we are computing combine_shares?  I am worried that this is a misuse resistance issue--this puts the onus on the programmer to remember to call verify_share when the shares are received.

Maybe the default should be to automatically verify all of the shares, although we could have an option to turn this off?

ursa_sharing/src/lib.rs
```diff
@@ -1 +1,175 @@
...
+//! one versus the other. Indeed both are used in
+//! <http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.134.6445&rep=rep1&type=pdf>.
+//!
+//! Feldman reveals the public value of the verifier whereas Pedersen's hides it.
+//!
+//! FUTURE: Adept secret sharing as described by Phillip Rogaway
+//! (see <https://eprint.iacr.org/2020/800>
+//!
+//! Future work would be to use reed-solomon
+//! codes to check for corrupted shares.
```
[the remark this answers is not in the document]
mikelodder7: Shouldn't need to be. Reed-solomon is just an encoding for fixing corrupted data.

ursa_sharing/src/pedersen.rs
mikelodder7: It's not terribly time-consuming to call verify_share for each one but I did want to benchmark them separately. Perhaps it's worth the overhead.

ursa_sharing/src/shamir.rs
```diff
@@ -0,0 +1,205 @@
...
+use rand::{CryptoRng, RngCore};
+use std::{collections::BTreeSet, convert::TryFrom};
+use zeroize::Zeroize;
+
+/// A Shamir simple secret share
+/// provides no integrity checking
+#[derive(Debug, Zeroize)]
+#[zeroize(drop)]
+pub struct Share {
+    pub(crate) identifier: usize,
```
dcmiddle, line 27: You may want u32 rather than usize. 
usize is architecture dependent but below it is assumed to be 4 bytes. 
There's probably no benefit to 32 to 64 bit flexibility here since 32 bit will still give you 4 billion shares. (And frankly if most of earth has to simultaneously agree to something the aliens have already won.)

dcmiddle (changes requested):
I reviewed lib.rs, shamir, bn3072, and some of the other general pieces. I did not closely review feldman or pedersen or the example curves.
It would be great if other reviewers could tackle those pieces.

hartm:
I'm happy with the state this is in, but can you deal with Dan's changes?  Thanks!

ursa_sharing/Cargo.toml
```diff
@@ -1,10 +1,39 @@
...
+[dev-dependencies]
+curve25519-dalek = "3.0"
+digest = { version = "0.8", features = ["std"] }
+ff = "0.8"
+ff-zeroize = "0.6"
+k256 = "0.5"
+openssl = "0.10"
+p256 = "0.5"
+pairing-plus = "0.19"
+sha2 = "0.8"
```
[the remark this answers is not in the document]
mikelodder7: pairing-plus is bound to sha0.8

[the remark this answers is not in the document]
mikelodder7: Luckily its just for the example code. Secret sharing doesn't use curves.

hartm (approved):
Thanks Mike!  LGTM.  All my comments have been addressed.

dcmiddle (approved):

PR #170, opened 2021-01-29
add pairings library

mikelodder7:
Add pairings library which brings `pairing-plus` up to date with newer RustCrypto crates and fixes the hash-to-curve implementation to match the latest IETF spec.

The reason to bring it in here is due to the original repo not accepting the PR to update their repo and failure to maintain the repo. Luckily, there isn't much that is going to change with this so we passively maintain this library for Ursa's other primitives that depend on it.


dcmiddle:
What do you think about forking that library and maintaining the patch in the fork? That seems to be the most common practice when I've looked into handling unmaintained dependencies. It preserves all of the existing boundaries and if the upstream project resumes maintenance then we can easily point back to it again.

mikelodder7:
I’m fine with that

hartm:
@dcmiddle , where should the fork live then?  Happy to defer to your open-source experience here.

dcmiddle:
> @dcmiddle , where should the fork live then? Happy to defer to your open-source experience here.

Could live under mike's github account like mikelodder7/pairings or under HL like hyperledger/ursa-pairings.

mikelodder7:
I’m happy to host it under my account

mac-arrap:
@mikelodder7 should we close this PR or are we still needing this change in the code base?

brentzundel:
This is a very old PR that has not seen any activity in months. 
If no opposition to doing so is raised in the next week, I am going to close this PR.

PR #176, opened 2021-03-20
Avoid reallocation when creating BBS public key

andrewwhitehead:
Currently the vec is one byte short, so it reallocates from 201 to 402 bytes.


dcmiddle (approved):

hartm:
Oh, that's a nice catch.

What kind of impact are you seeing on performance?

hartm (approved):

andrewwhitehead:
@hartm I just ran quick benchmark but I don't think the difference is measurable given the other work the function is doing. The BBS vs PS benchmarks in libzmix don't compile at the moment.

I just added another commit that fixes a similar issue in BlindSignature::new.

dhuseby:
I'll go ahead and try to fix and merge this.

PR #171, opened 2021-01-29
Migrate to k256

dcmiddle:
Separate commits for some namespace/dependency clarity and then the migration itself in its own commit.

Issue #151, opened 2020-10-05
Operation Oso


mikelodder7:
As part of the Ursa refactor, I'm breaking down the various tasks that anyone can do to contribute. Some of these tasks will need to be completed in order but may will not. The list is included here as a master list but each task will be detailed in an issue. The main goal here is to leave the current APIs as is or make small modifications and leave the old existing stuff for the moment until this operation is completed. At that point, we can start removing legacy code from the main project.

- [x] Create the top-level project library - #152 
- [ ] Create a subproject for core components - #153 
- [ ] Create a subproject for signatures
- [ ] Create a subproject for key agreement
- [ ] Create a subproject for encryption
- [x] Create a subproject for secret sharing
- [ ] Create a subproject for short group signatures
- [ ] Create a subproject for accumulators
- [ ] Create a subproject for circuits
- [ ] Refactor the current [signature](https://github.com/hyperledger/ursa/tree/master/libursa/src/signatures) folder to be the signatures subproject
- [ ] Refactor the current [encryption](https://github.com/hyperledger/ursa/tree/master/libursa/src/encryption) to be the encryption subproject
- [ ] Refactor the current [kex](https://github.com/hyperledger/ursa/tree/master/libursa/src/kex) to be the key agreement
- [ ] Refactor the current [sharing](https://github.com/hyperledger/ursa/tree/master/libursa/src/sharing) to be the sharing subproject
- [ ] Refactor [bbs](https://github.com/hyperledger/ursa/tree/master/libzmix/bbs) and [ps](https://github.com/hyperledger/ursa/tree/master/libzmix/src/signatures) to be the short group signatures subproject
- [ ] Refactor [bulletproofs](https://github.com/hyperledger/ursa/tree/master/libzmix/bulletproofs_amcl) to be part of the circuits subproject

dcmiddle:
@mikelodder7 Do I read this correctly that the zmix directory goes away in this reorganization? (I'm good with that, just checking my understanding of this layout).

mikelodder7:
Yes I believe it’s unnecessary

hartm:
> @mikelodder7 Do I read this correctly that the zmix directory goes away in this reorganization? (I'm good with that, just checking my understanding of this layout).

[...]

PR #180, opened 2021-05-14
Run libursa tests on Ubuntu 20.04

nhwn:
This PR adds `libursa`-specific jobs to GH Actions (namely building `libursa` and running its tests).

One thing I'm not sure about is the workaround for getting `libursa` to build on its own -- currently, running `cargo build` in the `libursa` directory on the `main` branch will yield this error:
```
error: current package believes it's in a workspace when it's not:
current:   ursa/libursa/Cargo.toml
workspace: ursa/Cargo.toml

this may be fixable by adding `libursa` to the `workspace.members` array of the manifest located at: ursa/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```
I heeded the second suggestion and excluded `libzmix` and `libursa` from the workspace, but I'm not sure if this is the correct course of action. Feel free to request/make changes!

As for the changes to `libursa/src/kex/secp256k1.rs`, the compatibility test was never updated after `libsecp256k1` was replaced with `k256` in #171, so this PR also fixes that.

### DIFF
--- a/libursa/src/kex/secp256k1.rs
+++ b/libursa/src/kex/secp256k1.rs
@@ -228,20 +228,19 @@ mod tests {
 
     #[test]
     fn secp256k1_compatibility() {
-        use libsecp256k1::{
-            ecdh::SharedSecret,
-            key::{PublicKey, SecretKey},
+          use k256::{
+            {PublicKey, SecretKey},
+            elliptic_curve::ecdh::diffie_hellman,
         };
 
         let scheme = EcdhSecp256k1Sha256::new();
         let (pk, sk) = scheme.keypair(None).unwrap();
-
-        let sk1 = SecretKey::from_slice(&sk[..]).unwrap();
-        let pk1 = PublicKey::from_slice(&pk[..]).unwrap();
-        let secret = SharedSecret::new(&pk1, &sk1);
+        let sk1 = SecretKey::from_bytes(&sk[..]).unwrap();
+        let pk1 = PublicKey::from_sec1_bytes(&pk[..]).unwrap();
+        let secret = diffie_hellman(sk1.secret_scalar(), pk1.as_affine());
         assert_eq!(
-            secret.as_ref(),
-            scheme.compute_shared_secret(&sk, &pk).unwrap().as_ref()
+            secret.as_bytes().to_vec(),
+            scheme.compute_shared_secret(&sk, &pk).unwrap().as_ref().to_vec()
         );
     }
 }
--- a/libursa/src/kex/secp256k1.rs
+++ b/libursa/src/kex/secp256k1.rs
@@ -130,8 +130,6 @@ mod ecdh_secp256k1 {
     use sha2::digest::generic_array::typenum::U32;
     use sha2::Digest;
     use zeroize::Zeroize;
-    #[allow(dead_code)]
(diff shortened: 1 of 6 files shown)


dcmiddle:
This is great. Thanks for catching & fixing the compatibility test.

For the workspace, I think the endgame is that libzmix is gone. I *think* Mike's intent is that libursa gets split up into the other components, like ursa_signatures so libursa goes away too.

Until that's all refactored, maybe we should just put libursa into the workspace members list so it gets built and tested?

libursa/src/kex/secp256k1.rs
```diff
@@ -130,6 +130,7 @@ mod ecdh_secp256k1 {
     use sha2::digest::generic_array::typenum::U32;
     use sha2::Digest;
     use zeroize::Zeroize;
+    #[allow(dead_code)]
```
dcmiddle, line 133: Probably better to remove the dead code. What was coming up?

nhwn: It's not used anywhere, I was just taking a super conservative approach for my changes. I'll go ahead and remove it now.

nhwn:
If we want to put `libursa` in the workspace members, I think the package name listed in the top-level `Cargo.toml` needs to change to something other than `ursa` (otherwise, we'll have 2 packages named `ursa` in the workspace, which gets rejected by `cargo`). Honestly, I'm not quite sure what the top-level crate is for since it only re-exports `ursa_sharing` at the moment, so I don't know if it's risky to rename it to something else or not.

Also, adding `libursa` to the workspace members won't actually _build_ `libursa` unless it's listed as a dependency in the top-level crate (this is in context of running a plain old `cargo build` in the top-level directory), Again, I'm unfamiliar with what the top-level crate is supposed to do, so I'll need further input here.

As for actually getting `libursa`'s tests to run in this configuration, we'll need to change the job's `cargo` invocation to either directly test the `libursa` package (e.g. `cargo test -p ursa`) or test _all_ the packages in the workspace (e.g. `cargo test --workspace`) because `cargo test` will only run the tests in the top-level crate (and there are none).

Do we want to test all the packages simultaneously, or do we want some sort of granularity? (to be clear, I think having everything in the workspace makes sense -- I just want to make sure these changes don't get in the way later)

mac-arrap (approved):
I have nothing to add more than @dcmiddle  has mentioned. Thank you for contributing this

mac-arrap:
@mikelodder7 Do you have any input for the concerns listed above?

dcmiddle:
Based on the conversation in today's community meeting we should rely on the Operation Oso issue description. Essentially don't "fix" libursa in the workspace and CI. Instead to get those components back into testing they should be refactored into the new structure e.g. signatures subproject.
https://github.com/hyperledger/ursa/issues/151

The fixes for k256k1 etc. should still remain in this PR though (recognizing that it's awkward that those changes aren't validated in CI until the refactoring happens)

nhwn:
I've reverted the changes to the workspace/CI as per what we discussed in Wednesday's meeting. At this point, this PR just fixes that one regression test and some compiler warnings, so I'll be opening up more PRs geared at #151 to follow up on this one.

brentzundel (approved):

dcmiddle (approved):

mac-arrap (approved):

PR #197, opened 2021-12-14
Update Time crate

dcmiddle:
Fixes https://github.com/hyperledger/ursa/issues/196

PR #198, opened 2022-01-05
Fix 'cl' feature, warnings

andrewwhitehead:
Compilation currently fails when the `cl` feature is enabled due to the `digest` dependency being updated previously.

Also fixed some warnings and skipped tests around the optional `serde` feature and non-existent `serialization` feature.


dcmiddle (approved):

brentzundel (approved):

andrewwhitehead:
The current audit failure is due to an CVE for the `time` crate, that version is being brought in as a transitive dev-dependency of `libsodium-ffi` (which incidentally hasn't been updated in a few years). I think [this PR](https://github.com/zip-rs/zip/pull/254) might fix the issue, or else it could be manually ignored.

dcmiddle:
yep .. see also https://github.com/hyperledger/ursa/pull/197 which is also blocked on the zip / time issue.

Issue #196, opened 2021-12-06
Security vulnerability with `time: v0.1.43`


appetrosyan:
Hi, we're using ursa in hyperledger iroha, When we ran `cargo audit` We found 
```
Crate:         time
Version:       0.1.43
Title:         Potential segfault in the time crate
Date:          2020-11-18
ID:            RUSTSEC-2020-0071
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:      Upgrade to >=0.2.23
Dependency tree: 
time 0.1.43
```

And `cargo tree` showed `time` as a direct dependency of `ursa`.

hartm:
I think time is only used for performance benchmarking, so this shouldn't be a security vulnerability for Ursa.  But it definitely should be updated, so thanks for pointing this out!

PR #200, opened 2022-01-06
Remove libsodium and time dependencies

andrewwhitehead:


andrewwhitehead:
NB I haven't built and tested the docker images.

mac-arrap:
We discussed during the last URSA meeting that we wanted to switch from libsodium to Dalek. This seems to just remove where libsodium is called but doesn't replace it.

andrewwhitehead:
libsodium wasn't used for ed25519 except in tests (verifying the Dalek backend) and in benchmarks.

dcmiddle (approved):
thanks! looks good (just one propagated spelling error)

dcmiddle:
> Somewhat related but I don't think this repo should have a Cargo.lock. It causes older dependency versions to be pinned when running tests (and the audit) where they would be updated to the latest patch versions when the library is used.

I don't like the lock file in here either. Official guidance is that libraries should not commit the lock file[1] unless the library is a static or cdylib[2]. Ursa is setup as a static, cdylib[3] ... but why? I'm not sure. If you want to remove it, I'm supportive but please create a separate issue so we can make sure we understand why ursa is setup this way.
 
[1]https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
[2]https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
[3]https://github.com/hyperledger/ursa/blob/main/libursa/Cargo.toml#L35

dcmiddle:
fixes #196

mac-arrap:
If it's agreed that this benchmarking can be removed then I see no problem.  The only reason I stated the above was per the request given to me on the last URSA call. But I do agree that it's not useful to just test another libraries backend.  @hartm  please review

mac-arrap (approved):

PR #211, opened 2022-08-04
Cleanup

blu3beri:
Noticed that there were quite some unnecessary references and small things that could be refactored. If these references and other choices are by-design This PR can be closed and ignored.

It does not alter the functionality one bit except for what seemd to me as an issue in `to_bytes` on the `Sharing` struct.

from:

```rust
    pub fn to_bytes(&self) -> Vec<u8> {
         let mut o = self.identifier.to_be_bytes().to_vec();
         o.append(&mut self.to_bytes()); // This recusively calls itself and should never return?
         o
     }
```

to:

```rust
    pub fn to_bytes(&mut self) -> Vec<u8> {
         let mut o = self.identifier.to_be_bytes().to_vec();
         o.append(&mut self.value);
         o
     }
```


brentzundel (approved):
These changes seem fine to me, as the tests are all still passing, but I'd appreciate if @mikelodder7 would weigh in.

mikelodder7 (approved):
LGTM

PR #216, opened 2022-11-24
Added Github Actions pipeline

blu3beri:
- Runs a CI pipeline for incoming PRs and merges to main.
  - audit
  - build
  - check
  - clippy
  - docs
  - format
  - tests
- Removed some files that I assumed are not used (please give this a careful review as I am not fully aware of the history of these files).
- Does not include the release to crates.io just yet.
- Moved the `ursa_` packages to an Ursa folder to keep the root a bit cleaner. They kept the `ursa_` prefix as we might want to release them as seperate packages, although some of them are empty.


brentzundel:
Is this still a draft?

blu3beri:
I think I left it in a draft as it still misses the release to crates.io. I also wasn't sure, or able to find,  what the continuous integration from appveyor is doing exactly. I think a couple of small additions, making releases possible, would complete the PR.

PR #217, opened 2023-01-17
fix(wasm): compilable and updated some of the bindings

blu3beri:
Did some local testing and ed25519sha512 and bls seem to work.

```typescript
import { Ed25519Sha512 } from "ursa";
import assert from "assert";

// Test vectors from ursa/libursa/src/signatures/ed25519.rs

const message = new TextEncoder().encode(
  "This is a dummy message for use with tests"
);

const signature = Buffer.from(
  "451b5b8e8725321541954997781de51f4142e4a56bab68d24f6a6b92615de5eefb74134138315859a32c7cf5fe5a488bc545e2e08e5eedfd1fb10188d532d808",
  "hex"
);

const publicKey = Buffer.from(
  "27c96646f2d4632d4fc241f84cbc427fbc3ecaa95becba55088d6c7b81fc5bbf",
  "hex"
);

assert(Ed25519Sha512.verify(message, signature, publicKey));
```

bls has some issues, maybe intentional, with dropping values (see g1 and g2).
```typescript
import assert from "assert";
import { Bls, Generator, SignKey, VerKey } from "ursa";

const msg = new Uint8Array([1, 1, 1]);

const seed = new Uint8Array(32);
const b = new Generator().toBytes();

// Creating of the same two generators as `Bls.sign` drops one and `Bls.verify` as well.
const g1 = Generator.fromBytes(b);
const g2 = Generator.fromBytes(b);

const signKey = SignKey.fromSeed(seed);
const signKey2 = SignKey.fromSeed(seed);

const verKey = new VerKey(g1, signKey);
const signature = Bls.sign(msg, signKey2);

assert(Bls.verify(msg, signature, verKey, g2));
```

Signed-off-by: blu3beri <blu3beri@proton.me>


blu3beri:
Also, I was curious what the intention of the `wasmprivatekey` and `wasmpublickey` were. I assume that you do not deal with the keys themselves in the JS side.

libursa/src/wasm/bls.rs
```diff
@@ -8,14 +8,17 @@ pub struct Generator(bls::Generator);
...
 #[wasm_bindgen]
 impl Generator {
+    #[wasm_bindgen(constructor)]
     pub fn new() -> Result<Generator, JsValue> {
         Ok(Generator(maperr!(bls::Generator::new())))
     }
 
-    pub fn as_bytes(&self) -> Vec<u8> {
+    #[wasm_bindgen(js_name = toBytes)]
+    pub fn to_bytes(&self) -> Vec<u8> {
```
brentzundel, line 17: I would expect changing the name from `as_bytes` to `to_bytes` would cause problems elsewhere, is that not the case?

brentzundel (commented):
thank you for the PR. I am curious about the choice to change the function name from `as_bytes` to `to_bytes`.

libursa/src/wasm/bls.rs
blu3beri: It would change it to a breaking change if the WASM module is used anywhere. Internally it still uses `as_bytes().to_vec()`.

I mainly changed it to be more inline with the JS way of doing a transformation. We do not really have an `as` way of doing it like this. 

TBH, I would be happy to change it back to be more inline with the rust interfaces.

brentzundel: If it doesn't break anything, I'm not bothered. Thanks for clarifying.

brentzundel (approved):

libursa/src/wasm/secp256k1.rs
```diff
@@ -83,8 +89,8 @@ impl EcdsaSecp256k1Sha256 {
 pub struct EcdhSecp256k1Sha256(EcdhSecp256k1Sha256Impl);
 
 #[wasm_bindgen]
-#[allow(non_snake_case)]
 impl EcdhSecp256k1Sha256 {
+    #[wasm_bindgen(constructor)]
     pub fn new() -> EcdhSecp256k1Sha256 {
```
appetrosyan, line 94:
```suggestion
    pub fn new() -> Self {
```

libursa/src/wasm/secp256k1.rs
```diff
@@ -58,22 +60,26 @@ impl EcdsaSecp256k1Sha256 {
...
 
-    pub fn publicKeyUncompressed(&self, pk: &WasmPublicKey) -> WasmPublicKey {
+    #[wasm_bindgen(js_name = publicKeyUnCompressed)]
+    pub fn public_key_uncompressed(&self, pk: &WasmPublicKey) -> WasmPublicKey {
         let pk = PublicKey::from(pk);
         WasmPublicKey::from(self.0.public_key_uncompressed(&pk))
     }
 
-    pub fn parseToPublicKey(&self, bytes: &[u8]) -> Result<WasmPublicKey, JsValue> {
+    #[wasm_bindgen(js_name = parse_to_public_key)]
```
appetrosyan, line 81: Was the meant to be camelCase.

libursa/src/wasm/cl.rs
```diff
@@ -344,15 +361,18 @@ pub struct ProofBuilder(cl::prover::ProofBuilder);
 #[wasm_bindgen]
 #[allow(non_snake_case)]
```
appetrosyan, line 362: Is this necessary?

libursa/src/wasm/cl.rs
```diff
@@ -39,38 +45,39 @@ impl NonCredentialSchema {
...
-//    }
-//}
+#[wasm_bindgen]
+impl CredentialValues {
+    #[wasm_bindgen(constructor)]
+    pub fn new() -> CredentialValues {
+        CredentialValues(cl::CredentialValuesBuilder::new().unwrap())
+    }
+
+    pub fn add_master_secret(&mut self, value: &MasterSecret) -> Result<(), JsValue> {
```
appetrosyan, line 55: Are these functions intentionally not renamed to camelCase?

appetrosyan:
In principle it looks good. Let me know once you've addressed the comments, and I'll approve.

blu3beri:
> In principle it looks good. Let me know once you've addressed the comments, and I'll approve.

All comments should be resolved.

appetrosyan (approved):

PR #222, opened 2023-02-06
updates to core

mikelodder7:
These updates define an error type that can be used across other packages that ursa makes available.


ursa_core/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
appetrosyan, line 1: Shared objects are a murky area. I do think that given that Ursa is a primarily going to be used as an rlib, this is the right call.

ursa_core/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+    CborError(#[from] serde_cbor::Error),
+    /// Convert bare errors
+    #[error("bare error")]
+    BareError(#[from] serde_bare::error::Error),
+    /// Convert thread access errors
+    #[error("thread access error")]
+    ThreadAccessError(#[from] std::thread::AccessError),
+    /// Generic errors to handle anything that implements the std::error::Error trait
+    #[error("error: {0}")]
+    Kind(Box<dyn std::error::Error>),
```
appetrosyan, line 33: I would personally try and avoid this. Instead it might be worth marking the entire structure as ```non_exhaustive```.

ursa_core/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn from(value: &String) -> Self {
+        Self::Kind(into_err(value.clone()))
+    }
+}
+
+impl From<&str> for UrsaError {
+    fn from(value: &str) -> Self {
+        Self::Kind(new_err(value))
+    }
+}
```
appetrosyan, line 52: This is odd. Usually one defines the impls in terms of `AsRef<str>`. Why have you decided to implement the conversion from String, instead of directly using &str?

ursa_core/src/lib.rs
```diff
@@ -1 +1,25 @@
+//!
```
appetrosyan, line 1: Missing documentation. This will not be overridden. I suggest adding in a couple of lines or removing the doc marker entirely.

ursa_core/src/lib.rs
```diff
@@ -1 +1,25 @@
...
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+mod error;
+
+pub use serde;
+pub use serde_bare;
+pub use serde_cbor;
+pub use serde_json;
+
+/// The errors generated by ursa components
```
appetrosyan, line 23: I would prefer it be documented in `error.rs`. That way the person re-exporting the module doesn't also have to document it.

ursa_core/src/error.rs
mikelodder7: Agreed. Done.

ursa_core/src/error.rs
mikelodder7: These are convenience methods to convert from a String. We don't want there to be a `core` update each time a new error is needed, thus the `Kind` value which can be anything that is derived from std::error::Error. The most common needed value is `String`.

ursa_core/src/lib.rs
mikelodder7: The values are documented. You also document each sub module. Not sure what you mean by your comment. Its common to have a short comment per sub module.

ursa_core/src/error.rs
appetrosyan: In that case, it might be worth considering `impl FromStr for Error`. This would also allow you to retain some strong typing via runtime checks. 

Additionally, these methods have a few issues. 

1. It's more common to have `impl From<AsRef<str>>`, which would allow you to instantiate it for more types that act like strings. 

2. `From<&String>` is allocating implicitly. Moreover, it would be covered by the `AsRef` impl, so you don't lose much by having the standard implementation. 

3. These functions are all repeated with no good reason. If I were to keep them, I'd perhaps delegate to the `&str` implementation. DRY.

ursa_core/src/lib.rs
appetrosyan: Suppose I decided to re-export the submodule from another place. I'd have to copy this `mod` declaration with its doc-comment. If instead `errors.rs` included ```//! Errors generated by Ursa components``` it would be less cumbersome. 

Here it doesn't make much difference, but in some situations it can save you a lot of hassle, and is less error prone.

ursa_core/src/error.rs
mikelodder7: To dry this out would be better for a macro. I've had unintended consequences happen when I've done `from<impl Trait> for ...`.

brentzundel (approved):

appetrosyan (approved):

PR #233, opened 2023-04-24
Move to EOL

mikelodder7:
I propose moving Ursa to EOL status.

None of the maintainers have the time or bandwidth to work on it anymore.
For any upstream projects, I'm happy to help migrate away from it within reason.


appetrosyan:
I vote NO. I have time to work on it, just not in an official capacity, and due to internal issues which I'd rather not make public. 

However, the project is useful to us, and while we could absorb it directly and turn it into `iroha_crypto` (it would make our review process easier), I do think that keeping the project alive a while longer makes sense.

swcurran:
I think this is premature and should not be done without considerably more thought on the impact of the change.  And we definitely don’t want to do while the dependencies on Ursa remain.

@mikelodder7 — have you discussed this with others?

mikelodder7:
Yes we discussed on maintainers call today

brentzundel:
I support moving Ursa to EOL unless it gets more active contribution.

I deeply appreciate the work that @appetrosyan has put in recently, but Mike and I don't have the capacity to contribute much in terms of code development. And few others have been contributing much of anything lately.

Others who have dependencies on Ursa will need to either pull those dependencies directly into their projects so EOL won't affect them, or contribute directly to Ursa so it can stay alive.

appetrosyan:
> Yes we discussed on maintainers call today

had there not been a mixup with calendars (because of which I assumed the call is on Wednesday), I would have provided you with some updates.

blu3beri:
I would also like to volunteer with maintaining the library from a non-cryptographic point of view, updating the CI, reviewing PRs, separation, etc if that helps.

danintel (approved):

tkuhrt (approved):

.github/settings.yml
```diff
@@ -4,14 +4,14 @@
 
 repository:
   name: ursa
-  description: Hyperledger Ursa is a shared cryptography library
+  description: Hyperledger Ursa has moved to end-of-life status.
```
swcurran, line 7: Request change to:

Hyperledger Ursa (a shared cryptographic library) has moved to end-of-life status, with the components of Ursa still in use moved to their relevant Hyperledger projects (AnonCreds, Indy, Aries and Iroha).

README.md
```diff
@@ -1,8 +1,6 @@
 # HYPERLEDGER URSA
 
-[![Ursa CI](https://github.com/hyperledger/ursa/workflows/Ursa%20CI/badge.svg)](https://github.com/hyperledger/ursa/actions?query=workflow%3A%22Ursa+CI%22)
-[![Build status](https://ci.appveyor.com/api/projects/status/1htmp82mdvmfjjap?svg=true)](https://ci.appveyor.com/project/hyperledger/ursa)
-[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/2447/badge)](https://bestpractices.coreinfrastructure.org/projects/2447)
+[Hyperledger Ursa was moved to EOL at the TOC meeting on 27 APR 2023.](https://toc.hyperledger.org/meeting-minutes/2023/2023-04-27-TOC-meeting-record.html)
```
swcurran, line 3: Request the following being added.  @appetrosyan — not sure if you want to add a reference to Iroha here.

```
The components of Ursa that are still in active use are moved to their respective projects where they will continue to be maintained, including:

- CL Signatures to AnonCreds
- BBS+ Signatures to Aries
- BLS Signatures to Indy

The published Ursa crates remain available for those projects already using them.
```

swcurran (changes requested):

.github/settings.yml
ryjones: updated

README.md
ryjones: updated

swcurran (approved):

